PR #10, opened 2022-01-22
Update to match changes in casper-node due to fast-sync feature

Fraser999:
This PR makes the `feat-fast-sync` branch compatible with the same branch of `casper-node`.

Many of the JSON-specific types have been copied here as the first step to breaking the dependency of the client on the node library.

In the near future, an approach should be settled to allow testing that the copied types stay in sync.

### DIFF
--- a/lib/validation.rs
+++ b/lib/validation.rs
@@ -12,18 +12,18 @@ use casper_node::{
         chain::{BlockIdentifier, EraSummary, GetEraInfoResult},
         state::GlobalStateIdentifier,
     },
-    types::{
-        json_compatibility, Block, BlockHeader, BlockValidationError, JsonBlock, JsonBlockHeader,
-    },
+    types::json_compatibility,
 };
 use casper_types::{bytesrepr, Key, StoredValue, U512};
+
+use crate::types::{self, Block, BlockHash, BlockHeader};
 
 const GET_ITEM_RESULT_BALANCE_VALUE: &str = "balance_value";
 const GET_ITEM_RESULT_STORED_VALUE: &str = "stored_value";
 const GET_ITEM_RESULT_MERKLE_PROOF: &str = "merkle_proof";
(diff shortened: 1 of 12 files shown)


lib/validation.rs
```diff
@@ -12,18 +12,18 @@ use casper_node::{
...
         state::GlobalStateIdentifier,
     },
-    types::{
-        json_compatibility, Block, BlockHeader, BlockValidationError, JsonBlock, JsonBlockHeader,
-    },
+    types::json_compatibility,
 };
 use casper_types::{bytesrepr, Key, StoredValue, U512};
 
+use crate::types::{self, Block, BlockHash, BlockHeader};
```
fizyk20, line 19: Why duplicate all these types in client, instead of just using the `casper_types` ones? That's a lot of code duplication, which looks like it might cause out-of-sync issues in the future.

Fraser999: There are a couple of reasons.

The main one right now is that these types don't exist in the `casper-types` crate, but rather in the `casper-node` crate.  They should be moved soon though, so that point is only valid temporarily.

The second reason (which I think is a weaker point) is that the client will need to support multiple versions of a given type.  I believe there's a notion that importing multiple versions of the `casper-types` crate could be more confusing and possibly less ergonomic than just duplicating the various supported versions of the specific types upon which the client depends.

As the `casper-types` crate evolves, I'd be happy to revisit that decision.

fizyk20: > The main one right now is that these types don't exist in the casper-types crate, but rather in the casper-node crate.

Ah, good point, but... Aren't they exported from casper-node, then? It's already a dependency of the client, so if the types are public there, it should be possible to import them, right?

In any case, not really a blocker, I just wanted to learn about the rationale :+1:

TomVasile (approved):
checked against `feat-fast-sync` branch of casper-node

PR #35, opened 2022-06-30
Avoid passing null params in RPC requests

Fraser999:
This fixes the client to avoid sending JSON-RPC requests with `"params":null` since this is invalid and will be rejected by the node in the upcoming version.

Closes [#3152](https://github.com/casper-network/casper-node/issues/3152).


goral09 (approved):

georgepisaltu (approved):

rafal-ch (approved):
LGTM :+1:

PR #36, opened 2022-07-01
Remove crypto module in favour of the one now available in casper-types

Fraser999:
This PR fixes breakages caused by the recent upgrade to Rust 1.62.0.

The change is predominantly around the planned update to remove the copied `crypto` module, which was copied from `casper-node` before that code was moved to the `casper-types` crate.


rafal-ch (approved):
LGTM :+1:

PR #37, opened 2022-07-15
Update to handle the changed structure of `node_state` in a get-node-status response

Fraser999:
This PR updates the client in line with the changes introduced in https://github.com/casper-network/casper-node/pull/3215.


rafal-ch (approved):
LGTM :+1:

PR #39, opened 2022-08-26
Reuse `reqwest::Client` in JSON RPC call

georgepisaltu:
This PR refactors the usage of `reqwest::Client` to reuse the same structure whenever users make requests. This is feasible because the client pools requests asynchronously, so reusing it consumes fewer resources.

`once_cell`'s `OnceCell` was used instead of `std`'s one because it's still an unstable feature in `std` as of right now.

### DIFF
--- a/lib/json_rpc/call.rs
+++ b/lib/json_rpc/call.rs
@@ -7,6 +7,8 @@ use serde_json::json;
 use crate::{Error, JsonRpcId, SuccessResponse, Verbosity};
 
 const RPC_API_PATH: &str = "rpc";
+/// Statically declared client used when making HTTP requests
+/// so opened connections are pooled.
 static CLIENT: OnceCell<Client> = OnceCell::new();
 
 /// Struct representing a single JSON-RPC call to the casper node.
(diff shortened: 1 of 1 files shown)


lib/json_rpc/call.rs
```diff
@@ -1,11 +1,13 @@
 use jsonrpc_lite::{JsonRpc, Params};
+use once_cell::sync::OnceCell;
 use reqwest::Client;
 use serde::{de::DeserializeOwned, Serialize};
 use serde_json::json;
 
 use crate::{Error, JsonRpcId, SuccessResponse, Verbosity};
 
 const RPC_API_PATH: &str = "rpc";
+static CLIENT: OnceCell<Client> = OnceCell::new();
```
marc-casperlabs, line 10: This needs a docstring :)

marc-casperlabs (approved):
Only minor thing is the missing docstring :)

lib/json_rpc/call.rs
georgepisaltu: Indeed.

PR #44, opened 2022-12-07
Preparation for fast sync v2

Fraser999:
This PR brings over the changes already in `dev` along with [a final commit](https://github.com/casper-ecosystem/casper-client-rs/commit/6eff666a3580ac67905e360c56bc59034841c43a) to accommodate for the recent changes in the status endpoint.


rafal-ch (approved):

lib/rpcs/v1_5_0/get_node_status.rs
```diff
@@ -7,156 +7,55 @@ use casper_hashing::Digest;
...
-        /// proceeds).
-        #[serde(rename = "number_of_remaining_tries_to_fetch")]
-        num_tries_to_fetch: usize,
-    },
-    /// Currently fetching the block's signatures.
-    #[serde(rename = "block signatures")]
-    BlockSignatures,
+/// The state of the reactor.
+#[derive(Copy, Clone, PartialOrd, Ord, PartialEq, Eq, Hash, Serialize, Deserialize, Debug)]
+pub enum ReactorState {
```
georgepisaltu, line 14: Maybe we should consider making public all of the stuff from the node that we're using in other places, like `casper-db-utils` and this repo. This code duplication will prove hard to maintain when we change things.

georgepisaltu (approved):
LGTM

lib/rpcs/v1_5_0/get_node_status.rs
Fraser999: Yeah, couldn't agree more.  This already is on the cards for node 2.0.

Issue #57, opened 2023-03-23
Support speculative execution via the casper-client


RitaMAllenCA:
The `put-deploy`, `send-deploy `and `transfer` subcommands should all be extended to support sending the deploy to the node's new `speculative_exec` endpoint.  This should take the form of a new arg `--speculative-exec` for each subcommand which itself takes an optional value of a `BlockIdentifier`.

PR #62, opened 2023-03-29
Fixes for casper-client-rs #57

bpr:
Merge me!

### DIFF
--- a/lib/cli.rs
+++ b/lib/cli.rs
@@ -73,6 +73,7 @@ pub use simple_args::help as simple_args_help;
 pub async fn put_deploy(
     maybe_rpc_id: &str,
     node_address: &str,
+    maybe_block_identifier: &str,
     verbosity_level: u64,
     deploy_params: DeployStrParams<'_>,
     session_params: SessionStrParams<'_>,
@@ -81,9 +82,16 @@ pub async fn put_deploy(
     let rpc_id = parse::rpc_id(maybe_rpc_id);
     let verbosity = parse::verbosity(verbosity_level);
     let deploy = deploy::with_payment_and_session(deploy_params, payment_params, session_params)?;
-    crate::put_deploy(rpc_id, node_address, verbosity, deploy)
-        .await
-        .map_err(CliError::from)
+    let block_identifier = parse::block_identifier(maybe_block_identifier)?;
+    if block_identifier.is_some() {
+        crate::speculative_exec(rpc_id, node_address, block_identifier, verbosity, deploy)
+            .await
+            .map_err(CliError::from)
+    } else {
+        crate::put_deploy(rpc_id, node_address, verbosity, deploy)
+            .await
+            .map_err(CliError::from)
+    }
 }
 
 /// Creates a [`Deploy`] and outputs it to a file or stdout.
@@ -131,15 +139,23 @@ pub fn sign_deploy_file(
 pub async fn send_deploy_file(
     maybe_rpc_id: &str,
     node_address: &str,
+    maybe_block_identifier: &str,
     verbosity_level: u64,
     input_path: &str,
 ) -> Result<SuccessResponse<PutDeployResult>, CliError> {
     let rpc_id = parse::rpc_id(maybe_rpc_id);
+    let block_identifier = parse::block_identifier(maybe_block_identifier)?;
     let verbosity = parse::verbosity(verbosity_level);
     let deploy = crate::read_deploy_file(input_path)?;
-    crate::put_deploy(rpc_id, node_address, verbosity, deploy)
-        .await
-        .map_err(CliError::from)
+    if block_identifier.is_some() {
+        crate::speculative_exec(rpc_id, node_address, block_identifier, verbosity, deploy)
+            .await
+            .map_err(CliError::from)
+    } else {
+        crate::put_deploy(rpc_id, node_address, verbosity, deploy)
+            .await
+            .map_err(CliError::from)
+    }
 }
 
 /// Transfers funds between purses.
@@ -156,6 +172,7 @@ pub async fn send_deploy_file(
 pub async fn transfer(
     maybe_rpc_id: &str,
     node_address: &str,
+    maybe_block_identifier: &str,
     verbosity_level: u64,
     amount: &str,
     target_account: &str,
@@ -173,9 +190,16 @@ pub async fn transfer(
         deploy_params,
         payment_params,
     )?;
-    crate::put_deploy(rpc_id, node_address, verbosity, deploy)
-        .await
-        .map_err(CliError::from)
+    let block_identifier = parse::block_identifier(maybe_block_identifier)?;
+    if block_identifier.is_some() {
+        crate::speculative_exec(rpc_id, node_address, block_identifier, verbosity, deploy)
+            .await
+            .map_err(CliError::from)
+    } else {
+        crate::put_deploy(rpc_id, node_address, verbosity, deploy)
+            .await
+            .map_err(CliError::from)
+    }
 }
 
 /// Creates a transfer [`Deploy`] and outputs it to a file or stdout.
--- a/lib/cli.rs
+++ b/lib/cli.rs
@@ -86,12 +86,12 @@ pub async fn put_deploy(
         .map_err(CliError::from)
 }
 
-/// Speculatively creates a [`Deploy`] and sends it to the network for execution.
+/// Creates a [`Deploy`] and sends it to the specified node for speculative execution.
 ///
 /// For details of the parameters, see [the module docs](crate::cli#common-parameters) or the docs
 /// of the individual parameter types.
 pub async fn speculative_put_deploy(
-    maybe_speculative_exec: &str,
+    maybe_block_id: &str,
     maybe_rpc_id: &str,
     node_address: &str,
     verbosity_level: u64,
@@ -102,7 +102,7 @@ pub async fn speculative_put_deploy(
     let rpc_id = parse::rpc_id(maybe_rpc_id);
     let verbosity = parse::verbosity(verbosity_level);
     let deploy = deploy::with_payment_and_session(deploy_params, payment_params, session_params)?;
-    let speculative_exec = parse::block_identifier(maybe_speculative_exec)?;
+    let speculative_exec = parse::block_identifier(maybe_block_id)?;
     crate::speculative_exec(rpc_id, node_address, speculative_exec, verbosity, deploy)
         .await
         .map_err(CliError::from)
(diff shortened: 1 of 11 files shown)


lib/rpcs/v1_4_5/speculative_exec.rs
```diff
@@ -0,0 +1,35 @@
...
+
+/// The `result` field of a successful JSON-RPC response to an `speculative_exec` request.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct SpeculativeExecResult {
+    /// The JSON-RPC server version.
+    pub api_version: ProtocolVersion,
+    /// The deploy hash.
+    pub deploy_hash: DeployHash,
+}
```
Fraser999, line 35: This needs to be the same as the type defined by the node [here](https://github.com/casper-network/casper-node/blob/cc95de3a80cdada1d67a84a2178f23bcf10e36ae/node/src/components/rpc_server/rpcs/speculative_exec.rs#L58-L66):
```suggestion
pub struct SpeculativeExecResult {
    /// The JSON-RPC server version.
    pub api_version: ProtocolVersion,
    /// Hash of the block on top of which the deploy was executed.
    pub block_hash: BlockHash,
    /// Result of the execution.
    pub execution_result: ExecutionResult,
}
```

lib/cli.rs
```diff
@@ -73,6 +73,7 @@ pub use simple_args::help as simple_args_help;
 pub async fn put_deploy(
     maybe_rpc_id: &str,
     node_address: &str,
+    maybe_block_identifier: &str,
```
Fraser999, line 76: As with ordering the args in the subcommands, I'd probably try and keep the rpc_id, node_address and verbosity parameters grouping unbroken.

src/deploy/creation_common.rs
```diff
@@ -235,6 +236,33 @@ pub(super) fn payment_str_params(matches: &ArgMatches) -> PaymentStrParams<'_> {
...
+            .help(ARG_HELP)
+            .display_order(DisplayOrder::BlockIdentifier as usize)
+    }
+
+    pub(in crate::deploy) fn get(matches: &ArgMatches) -> &str {
+        matches
+            .get_one::<String>(ARG_NAME)
+            .map(String::as_str)
+            .unwrap_or_default()
+    }
```
Fraser999, line 263: I think we need to change the return type here to represent the three cases: `--speculative-exec` wasn't provided, `--speculative-exec` was provided with no value, and `--speculative-exec` was provided with a value:
```suggestion
    pub(in crate::deploy) fn get(matches: &ArgMatches) -> (bool, &str) {
        let value = matches
            .get_one::<String>(ARG_NAME)
            .map(String::as_str)
            .unwrap_or_default();
        match value {
            "" => (false, ""),
            DEFAULT_MISSING_VALUE => (true, ""),
            _ => (true, value),
        }
    }
```

lib/lib.rs
```diff
@@ -100,6 +101,26 @@ use types::{Deploy, DeployHash, MAX_SERIALIZED_SIZE_OF_DEPLOY};
...
+/// Sends a JSON-RPC `speculative_exec` request to the specified node.
+///
+/// For details of the parameters, see [the module docs](crate#common-parameters).
+pub async fn speculative_exec(
+    rpc_id: JsonRpcId,
+    node_address: &str,
+    maybe_block_identifier: Option<BlockIdentifier>,
+    verbosity: Verbosity,
+    deploy: Deploy,
+) -> Result<SuccessResponse<PutDeployResult>, Error> {
```
Fraser999, line 115: This needs to return a `SpeculativeExecResult` rather than a `PutDeployResult`.

As with other result types, it should be exported via `lib/rpcs/results.rs`.

lib/cli.rs
Fraser999, line 1: Since the speculative execution response is different to the put deploy response, we'll need to create three new functions here for this - something like `speculative_put_deploy`, `speculative_send_deploy_file` and `speculative_transfer` I guess.

src/deploy/put.rs
```diff
@@ -33,6 +34,7 @@ impl ClientCommand for PutDeploy {
         let verbosity_level = common::verbose::get(matches);
 
         let secret_key = common::secret_key::get(matches);
+        let maybe_block_identifier = creation_common::speculative_exec::get(matches);
```
Fraser999, line 37: With the other changes suggested, this will become `let (speculative_exec, maybe_block_identifier) =` and below we'll call `casper_client::cli::speculative_put_deploy` or `casper_client::cli::put_deploy` depending upon the bool.

Fraser999 (changes requested):

lib/lib.rs
```diff
@@ -100,6 +101,26 @@ use types::{Deploy, DeployHash, MAX_SERIALIZED_SIZE_OF_DEPLOY};
 pub use validation::ValidateResponseError;
 pub use verbosity::Verbosity;
 
+/// Puts a [`SpeculativeExec`] to the network for execution.
```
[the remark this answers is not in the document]
bpr: Done, yes much better.

lib/rpcs/v1_4_5/speculative_exec.rs
```diff
@@ -0,0 +1,35 @@
...
+impl SpeculativeExecParams {
+    pub(crate) fn new(block_identifier: Option<BlockIdentifier>, deploy: Deploy) -> Self {
+        SpeculativeExecParams {
+            block_identifier,
+            deploy,
+        }
+    }
+}
+
+/// The `result` field of a successful JSON-RPC response to an `speculative_exec` request.
```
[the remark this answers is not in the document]
bpr: Ha ha, the perils of copy pasta! Thanks!

lib/rpcs/v1_4_5/speculative_exec.rs
```diff
@@ -0,0 +1,35 @@
+use serde::{Deserialize, Serialize};
```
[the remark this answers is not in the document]
bpr: Fixed.

src/deploy/creation_common.rs
```diff
@@ -235,6 +236,33 @@ pub(super) fn payment_str_params(matches: &ArgMatches) -> PaymentStrParams<'_> {
...
 
+/// Handles providing the arg for speculative execution.
+pub(super) mod speculative_exec {
+    use super::*;
+
+    const ARG_NAME: &str = "speculative_exec";
+    const ARG_VALUE_NAME: &str = "HEX STRING OR INTEGER";
+    const ARG_HELP: &str =
+        "The endpoint to perform the speculative execution, identified by a BlockIdentifier, \
+        which is either a cryptographic hash or a u64.";
```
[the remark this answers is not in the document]
bpr: That's a pretty long help string. Almost two tweets!

src/deploy/creation_common.rs
```diff
@@ -235,6 +236,33 @@ pub(super) fn payment_str_params(matches: &ArgMatches) -> PaymentStrParams<'_> {
...
+
+    pub(in crate::deploy) fn arg() -> Arg {
+        Arg::new(ARG_NAME)
+            .long(ARG_NAME)
+            .required(false)
+            .value_name(ARG_VALUE_NAME)
```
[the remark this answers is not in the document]
bpr: `    const DEFAULT_MISSING_VALUE: &str = "0";
`

Will that cause problems?

lib/rpcs/v1_4_5/speculative_exec.rs
bpr: That would add a dependency on casper-node on account of BlockHash. What's the suggested workaround? IIRC, you don't want to introduce dependencies on node.

lib/cli.rs
bpr: I gotta say, if the grouping of parameters is that important, making them into a struct would make sense.

lib/lib.rs
bpr: Done.

src/deploy/put.rs
bpr: I took a slightly different but I hope, equivalent path, using an option type to encode the speculative_exec/no_speculative_exec distinction.

lib/cli.rs
```diff
@@ -86,6 +86,27 @@ pub async fn put_deploy(
         .map_err(CliError::from)
 }
 
+/// Speculatively creates a [`Deploy`] and sends it to the network for execution.
```
Fraser999, line 89: I think the term "speculative" applies exclusively to the execution of the deploy on the node receiving it.  It follows quite a different code path through the node, and a bunch of caveats are implied in any successful response.

So while I don't think this is the place to try and describe the node's process, I guess I'd rephrase this to something like:
```suggestion
/// Creates a [`Deploy`] and sends it to the specified node for speculative execution.
```

lib/cli.rs
```diff
@@ -142,6 +163,25 @@ pub async fn send_deploy_file(
         .map_err(CliError::from)
 }
 
+/// speculatively eads a previously-saved [`Deploy`] from a file and sends it to the network for execution.
```
Fraser999, line 166:
```suggestion
/// Reads a previously-saved [`Deploy`] from a file and sends it to the specified node for
/// speculative execution.
```

lib/cli.rs
```diff
@@ -178,6 +218,44 @@ pub async fn transfer(
         .map_err(CliError::from)
 }
 
+/// Speculatively transfers funds between purses.
```
Fraser999, line 221:
```suggestion
/// Creates a [`Deploy`] to transfer funds between purses, and sends it to the specified node for
/// speculative execution.
```

Fraser999 (changes requested):
Nearly there!  I made a few suggestions, but there's just one blocker left (the issue in `speculative_exec::get()` in creation_common.rs).

src/deploy/creation_common.rs
bpr: I don't understand why it doesn't work with

```Rust
    const DEFAULT_MISSING_VALUE: &str = "";
```

If that gets put in why isn't it `Some("")`?

Anyways, I changed as per your suggestion.

lib/cli.rs
```diff
@@ -86,6 +86,27 @@ pub async fn put_deploy(
         .map_err(CliError::from)
 }
 
+/// Speculatively creates a [`Deploy`] and sends it to the network for execution.
+///
+/// For details of the parameters, see [the module docs](crate::cli#common-parameters) or the docs
+/// of the individual parameter types.
+pub async fn speculative_put_deploy(
+    maybe_speculative_exec: &str,
```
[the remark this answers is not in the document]
bpr: Done

lib/cli.rs
bpr: Done

lib/cli.rs
bpr: Done

lib/cli.rs
bpr: Done.

Fraser999 (approved):
LGTM now, but I'd be happy for you to change the `speculative_exec::get()` function one last time as per the comment.  Up to you - it's functionally correct now - just more verbose than need be.

bpr:
What is an appropriate test for this PR?

darthsiroftardis (approved):

PR #70, opened 2023-05-04
Update era response validation logic

### LINKED ISSUE #69: Accommodate merkle proof in get_era_summary call
different versions of the block use different key schemes and we need to account for this in the new endpoint

darthsiroftardis:
CHANGELOG

- Update validation logic to account for the new Key::EraSummary

closes #69


lib/validation.rs
```diff
@@ -100,7 +100,11 @@ pub(crate) fn validate_get_era_info_response(
                 .map_err(|_| ValidateResponseError::ValidateResponseFailedToParse)?;
             let proofs: Vec<TrieMerkleProof<Key, StoredValue>> =
                 bytesrepr::deserialize(proof_bytes)?;
-            let key = Key::EraInfo(era_id);
+            let key = if result.api_version < ProtocolVersion::from_parts(1u32, 4u32, 14u32) {
```
Fraser999, line 103: This should work for our mainnet and testnet, but won't work for e.g. nctl networks being spun up from version 1.0.0.

Maybe we should try first using `Key::EraSummary` and fall back to `Key::EraInfo` if that fails?

darthsiroftardis: Added the change

Fraser999 (approved):

PR #78, opened 2023-06-20
Payment-amount is optional for both native transfers

### LINKED ISSUE #60: Make --payment-amount in client optional for both native transfer subcommands
In the Rust client, we should support no payment-related args being provided in the transfer and make-transfer subcommands. In the case that no payment-related args are provided by the user, the deploy's payment code should be constructed so that it results in a standard payment of 2_500_000_000 motes. For reference, see https://github.com/casper-network/casper-node/blob/91b760341b2dd93da87ac9b903f510ea952e2cc1/client/lib/parsing.rs#L524-L527 and https://github.com/casper-network/casper-node/blob/91b760341b2dd93da87ac9b903f510ea952e2cc1/client/lib/parsing.rs#L228-L239

We should also update the help strings output to a user of the client, and the rustdoc comments around the relevant lib code to make it clear that payment-related args are stored as part of the deploy, but are otherwise unused.

copied issued from https://app.zenhub.com/workspaces/engineering-l1-60953fafb1945f0011a3592d/issues/gh/casper-network/casper-node/2074

hoffmannjan:
Constant value of payment amount (`2500000000` motes) is now default value for both `transfer` and `make-transfer` subcommands.

When the `default_value` argument is not passed to `standard_payment_amount()` function it will automatically make `--payment-amount` required (so it works as before in `make-deploy` and `put-deploy`)

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -9,6 +9,7 @@ pub const ARG_PATH: &str = "PATH";
 pub const ARG_HEX_STRING: &str = "HEX STRING";
 pub const ARG_STRING: &str = "STRING";
 pub const ARG_INTEGER: &str = "INTEGER";
+pub const DEFAULT_TRANSFER_PAYMENT_AMOUNT: &'static str = "2500000000";
 
 /// Handles the arg for whether verbose output is required or not.
 pub mod verbose {
--- a/src/common.rs
+++ b/src/common.rs
@@ -9,7 +9,7 @@ pub const ARG_PATH: &str = "PATH";
 pub const ARG_HEX_STRING: &str = "HEX STRING";
 pub const ARG_STRING: &str = "STRING";
 pub const ARG_INTEGER: &str = "INTEGER";
-pub const DEFAULT_TRANSFER_PAYMENT_AMOUNT: &'static str = "2500000000";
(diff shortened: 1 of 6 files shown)


bpr (commented):
Are you sure you don't want to delare it `static` rather than `const`? IDK the answer to that question, BTW. IIRC `const` gets inlined and the `static` vars do not, and get their own memory. No idea if that matters here.

hoffmannjan:
> Are you sure you don't want to delare it `static` rather than `const`? IDK the answer to that question, BTW. IIRC `const` gets inlined and the `static` vars do not, and get their own memory. No idea if that matters here.

IDK the answer either, but I thought inlining it shouldn't cause any bad consequences. Looking [here](https://stackoverflow.com/a/52753798) and [at the RFC](https://github.com/rust-lang/rfcs/blob/master/text/0246-const-vs-static.md#summary) mentioned by the OP it makes sense for me - but as I said before I'm not 100% sure.

bpr:
I'd be very surprised if it caused any problems. The reason I mentioned it is that every time I see a static str constant it is declared like `static VARNAME: &str = "VARNAME";`  and the usage of `const` looked unusual.

bpr (approved):
I don't think the static vs const issue is worth holding this up, so....

LGTM

Fraser999 (approved):

PR #83, opened 2023-06-30
Removing wildcard dependencies to allow cargo publish.

sacherjj:
We cannot have '*' as dependency versions in Cargo.toml and publish to crates.io.  Replaced casper-hashing and casper-types.


sacherjj:
Going to try to re-publish 2.0.0 after this merges and see if we can get this out without incrementing version.

Fraser999 (approved):

PR #85, opened 2023-07-13
Initial update to match node 2.0 changes

### LINKED ISSUE #86: support for the changes made to the info_get_deploy RPC in node 2.0.
This PR adds support for the changes made to the info_get_deploy RPC in node 2.0.

As part of the node 2.0 work, many types upon which the client depends have been moved to casper-types from casper-node, meaning these no longer need to be duplicated in the client library.

One point to note is that validation of a get-block response has been disabled. There will likely be a further change made in casper-types which will allow the JsonBlock type to be removed in favour of the actual Block type, and Block provides a public verify() method which will be suitable for this use case. Rather than spending time refactoring the validation logic, I have left it commented out in anticipation of it being replaced soon.

A final point to note is that the casper-types dependency has a patch specified in the manifest of casper-types = { git = "https://github.com/Fraser999/casper-node", branch = "2124-use-execution-journal" }. Once the node PR using this fork and branch merges, we should update the patch to casper-types = { git = "https://github.com/casper-network/casper-node", branch = "feat-2.0" }.

Fraser999:
This PR adds support for the changes made to the `info_get_deploy` RPC in node 2.0.

As part of the node 2.0 work, many types upon which the client depends have been moved to `casper-types` from `casper-node`, meaning these no longer need to be duplicated in the client library.

One point to note is that validation of a `get-block` response has been disabled.  There will likely be a further change made in `casper-types` which will allow the `JsonBlock` type to be removed in favour of the actual `Block` type, and `Block` provides a public `verify()` method which will be suitable for this use case.  Rather than spending time refactoring the validation logic, I have left it commented out in anticipation of it being replaced soon.

A final point to note is that the `casper-types` dependency has a patch specified in the manifest of `casper-types = { git = "https://github.com/Fraser999/casper-node", branch = "2124-use-execution-journal" }`.  Once the node PR using this fork and branch merges, we should update the patch to `casper-types = { git = "https://github.com/casper-network/casper-node", branch = "feat-2.0" }`.

Closes #86.


bpr (approved):
LGTM

There were some comments  on commented code blocks during the walk thru, I believe the resolution was to put a `TODO` or `FIXME` in that region.

Fraser999:
> There were some comments on commented code blocks during the walk thru, I believe the resolution was to put a TODO or FIXME in that region.

Done in [aa2ee88](https://github.com/casper-ecosystem/casper-client-rs/pull/85/commits/aa2ee888d7d7a7bb7ac83671157309589ce6d7d7).

Issue #56, opened 2023-03-23
Improve multisig support in casper-client-rs


RitaMAllenCA:
We should support the creation of deploys where the session account is not automatically derived from the provided secret key.  We should also make the provision of a secret key optional for the `make-deploy` and `make-transfer` subcommands

Fraser999:
Note that updates to support this exist [here](https://github.com/casper-ecosystem/casper-client-rs/commit/45842b808d78259030651c6c80067012a2c6138a), but need the addition of tests before a PR can be cut.

RitaMAllenCA:
2.0.1

MParlikar:
This is a high value feature that has been rolling for several sprints. Needs to be prioritized and tied off.  Please work to get it done.

PR #91, opened 2023-07-27
Improve multisig support for make-deploy and make-transfer subcommands

zacshowa:
# Summary of work done:
- Changes to improve muiltisig support in the `make-deploy` and `make-transfer` subcommands
- Tests to verify that unsigned deploys and transfers can be created
- Tests to verify that attempts to create deploys or transfers with no associated account fail.
- Work to integrate above changes to a branch based on `dev` rather than `rustSDK`

# Associated ticket(s):
- #56 
- [Improve multisig support](https://app.zenhub.com/workspaces/engineering-l2-62618a6423afe100128e4656/issues/gh/casper-ecosystem/casper-client-rs/56)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,31 +14,7 @@ All notable changes to this project will be documented in this file.  The format
 ## Unreleased
 
 ### Added
-* Add module to support node 2.0.0 RPCs.
-
-### Changed
-* Update to match change to node RPC `info_get_deploy`.
-* Improve multisig support for `make-deploy`, and `make-transfer` subcommands.
-
-### Removed
-* Remove following public types which are now available in `casper_types`:
-  * `Account` and its related types `ActionThresholds` and `AssociatedKey`
-  * `Bid`
-  * `BidderAndBid`
-  * `Block` and its related types `BlockBody`, `BlockHash`, `BlockHashAndHeight` and `BlockHeader`
-  * `ChainspecRawBytes`
-  * `Contract`
-  * `ContractPackage` and its related types `ContractPackageStatus`, `ContractVersion`, `DisabledVersion` and `Group`
-  * `Delegator`
-  * `Deploy` and its related types `Approval`, `DeployBuilder`, `DeployHash` and `DeployHeader`
-  * `EraEnd` and its related types `EraReport`, `Reward` and `ValidatorWeight`
-  * `ExecutableDeployItem`
-  * `NamedKey`
-  * `Proof`
-  * `StoredValue`
-  * `TimeDiff`
-  * `Timestamp`
-  * `TransferTarget`
+* Add support for crafting unsigned deploys and transfers by providing an account, but not seccret key, to the `make-deploy` and `make-transfer` subcommands.
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,37 @@ All notable changes to this project will be documented in this file.  The format
 
 
 
+## Unreleased
+
+### Added
+* Add module to support node 2.0.0 RPCs.
+
+### Changed
+* Update to match change to node RPC `info_get_deploy`.
+* Improve multisig support for `make-deploy`, and `make-transfer` subcommands.
+
+### Removed
+* Remove following public types which are now available in `casper_types`:
+  * `Account` and its related types `ActionThresholds` and `AssociatedKey`
+  * `Bid`
+  * `BidderAndBid`
+  * `Block` and its related types `BlockBody`, `BlockHash`, `BlockHashAndHeight` and `BlockHeader`
+  * `ChainspecRawBytes`
+  * `Contract`
+  * `ContractPackage` and its related types `ContractPackageStatus`, `ContractVersion`, `DisabledVersion` and `Group`
+  * `Delegator`
+  * `Deploy` and its related types `Approval`, `DeployBuilder`, `DeployHash` and `DeployHeader`
+  * `EraEnd` and its related types `EraReport`, `Reward` and `ValidatorWeight`
+  * `ExecutableDeployItem`
+  * `NamedKey`
+  * `Proof`
+  * `StoredValue`
+  * `TimeDiff`
+  * `Timestamp`
+  * `TransferTarget`
+
+
+
 ## [2.0.0] - 2023-06-28
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ async-trait = "0.1.59"
 base16 = "0.2.1"
 base64 = "0.13.1"
 casper-hashing = "2.0.0"
-casper-types = { version = "3.0.0", features = ["std"] }
+casper-types = { version = "3.0.0", features = ["std", "json-schema"] }
 clap = { version = "4", features = ["cargo", "deprecated", "wrap_help"] }
 clap_complete = "4"
 derp = "0.0.14"
@@ -60,7 +60,7 @@ tokio = { version = "1.23.0", features = [
     "rt-multi-thread",
     "sync",
     "time",
-], optional = true }
+]}
 uint = "0.9.4"
 untrusted = "0.9.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,7 @@
 [package]
 name = "casper-client"
 version = "2.0.0" # when updating, also update 'html_root_url' in lib.rs
-authors = [
-    "Marc Brinkmann <marc@casperlabs.io>",
-    "Fraser Hutchison <fraser@casperlabs.io>",
-]
+authors = ["Marc Brinkmann <marc@casperlabs.io>", "Fraser Hutchison <fraser@casperlabs.io>",]
 edition = "2021"
 description = "A client library and binary for interacting with the Casper network"
 documentation = "https://docs.rs/casper-client"
@@ -37,12 +34,7 @@ hex-buffer-serde = "0.4.0"
 humantime = "2"
 itertools = "0.10.5"
 jsonrpc-lite = "0.6.0"
-k256 = { version = "0.7.3", features = [
-    "arithmetic",
-    "ecdsa",
-    "sha256",
-    "zeroize",
-] }
+k256 = { version = "0.7.3", features = ["arithmetic", "ecdsa", "sha256", "zeroize", ] }
 num-traits = "0.2.15"
 once_cell = "1"
 pem = "1"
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,37 @@ All notable changes to this project will be documented in this file.  The format
...
+  * `ExecutableDeployItem`
+  * `NamedKey`
+  * `Proof`
+  * `StoredValue`
+  * `TimeDiff`
+  * `Timestamp`
+  * `TransferTarget`
+
+
+
```
Fraser999: I don't think any of these changes relate to this PR.

There should be an entry under `Unreleased` though where you describe the new functionality added.

Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
 name = "casper-client"
 version = "2.0.0" # when updating, also update 'html_root_url' in lib.rs
-authors = ["Marc Brinkmann <marc@casperlabs.io>", "Fraser Hutchison <fraser@casperlabs.io>"]
+authors = [
+    "Marc Brinkmann <marc@casperlabs.io>",
+    "Fraser Hutchison <fraser@casperlabs.io>",
+]
```
Fraser999: I think we should be consistent about formatting toml arrays.  I'd suggest leaving this all on one line.  If you strongly prefer this style however, let's apply it to all the arrays.

lib/cli/deploy_str_params.rs
```diff
@@ -2,6 +2,9 @@
 #[derive(Default, Debug)]
 pub struct DeployStrParams<'a> {
     /// Path to secret key file.
+    ///
+    /// If `secret_key` is empty, the new deploy will not be signed and will need to be signed (e.g.
+    /// via [`sign_deploy_file`]) at least once in order to be made valid.
```
Fraser999:
```suggestion
    /// If `secret_key` is empty, the new deploy will not be signed and will need to be signed (e.g.
    /// via [`sign_deploy_file`](super::sign_deploy_file)) at least once in order to be made valid.
```

lib/cli/tests.rs
```diff
@@ -272,3 +294,82 @@ fn should_fail_to_create_transfer_with_bad_args() {
...
+        deploy_params_without_account(),
+        PaymentStrParams::with_amount("100"),
+        true,
+    );
+    assert!(transfer_deploy.is_err());
+    assert!(matches!(
+        transfer_deploy.unwrap_err(),
+        CliError::Core(Error::DeployMissingSessionAccount)
+    ));
+}
```
Fraser999, line 375: Maybe worth adding a couple more tests for checking that `deploy::with_payment_and_session` and `deploy::new_transfer` both return an error if their last arg is `false` and the deploy has no secret key?

zacshowa: When adding these tests I realized that despite returning an error in this case, some work needed to be done.
The error was related to the "key file"  (the empty string) being an invalid file location. I thought that this wasn't a great error message to return; The error should be more informative that the parameters were invalid.
I added an extra cause to the maybe_secret_key construction that does this check

```
 if ... {
     ...
} else if deploy_params.secret_key.is_empty() && !allow_unsigned_deploy {
        return Err(CliError::InvalidArgument {
            context: "with_payment_and_session (secret_key, allow_unsigned_deploy)",
            error: format!(
                "allow_unsigned_deploy was {}, but no secret key was provided",
                allow_unsigned_deploy
            ),
        });
} else {
    ...
}
```
If this isn't the right way to be more informative about this error case I am open to suggestions!

Cargo.toml
```diff
@@ -24,17 +24,17 @@ async-trait = "0.1.59"
 base16 = "0.2.1"
 base64 = "0.13.1"
 casper-hashing = "2.0.0"
-casper-types = { version = "3.0.0", features = ["std"] }
+casper-types = { version = "3.0.0", features = ["std", "json-schema"] }
```
Fraser999:
```suggestion
casper-types = { version = "3.0.0", features = ["std"] }
```

lib/cli/tests.rs
```diff
@@ -373,3 +372,51 @@ fn should_fail_to_create_transfer_without_account() {
...
+#[test]
+fn should_fail_to_create_transfer_with_no_secret_key_while_not_allowing_unsigned_deploy() {
+    let deploy_params = deploy_params_without_secret_key();
+    let payment_params =
+        PaymentStrParams::with_package_hash(PKG_HASH, VERSION, ENTRYPOINT, args_simple(), "", "");
+
+    let transfer_deploy = deploy::new_transfer(
+        "10000",
+        None,
+        "bad public key.",
```
Fraser999: I think it would be better to use a valid `PublicKey` here, as we ideally want just the one aspect to be invalid.

As currently written, this test would start failing if we re-order the arg parsing inside `new_transfer` to get `target` before getting `maybe_secret_key`.

lib/cli/deploy.rs
```diff
@@ -40,9 +61,22 @@ pub fn new_transfer(
...
+        return Err(CliError::InvalidArgument {
+            context: "new_transfer (secret_key, allow_unsigned_deploy)",
+            error: format!(
+                "allow_unsigned_deploy was {}, but no secret key was provided",
+                allow_unsigned_deploy
+            ),
+        });
+    } else {
+        Some(parse::secret_key_from_file(deploy_params.secret_key)?)
+    };
```
Fraser999, line 79: Nice change!

Fraser999 (commented):
~One~ Two nitpicks remaining.

lib/cli/deploy.rs
```diff
@@ -11,24 +11,45 @@ pub fn with_payment_and_session(
...
+        return Err(CliError::InvalidArgument {
+            context: "with_payment_and_session (secret_key, allow_unsigned_deploy)",
+            error: format!(
+                "allow_unsigned_deploy was {}, but no secret key was provided",
+                allow_unsigned_deploy
+            ),
+        });
+    } else {
+        Some(parse::secret_key_from_file(deploy_params.secret_key)?)
+    };
```
gRoussac, line 30:
```suggestion
    let maybe_secret_key = if !deploy_params.secret_key.is_empty() {
        Some(parse::secret_key_from_file(deploy_params.secret_key)?)
    } else {
        if !allow_unsigned_deploy {
            return Err(CliError::InvalidArgument {
                context: "with_payment_and_session (secret_key, allow_unsigned_deploy)",
                error: format!(
                    "allow_unsigned_deploy was {}, but no secret key was provided",
                    allow_unsigned_deploy
                ),
            });
        }
        None
    };
```

the if else is not very clear to read so I suggest you isolate your case first then decline the rest

lib/cli/deploy.rs
```diff
@@ -40,9 +61,22 @@ pub fn new_transfer(
...
+    } else if deploy_params.secret_key.is_empty() && !allow_unsigned_deploy {
+        return Err(CliError::InvalidArgument {
+            context: "new_transfer (secret_key, allow_unsigned_deploy)",
+            error: format!(
+                "allow_unsigned_deploy was {}, but no secret key was provided",
+                allow_unsigned_deploy
+            ),
+        });
+    } else {
+        Some(parse::secret_key_from_file(deploy_params.secret_key)?)
```
gRoussac, line 78: same comment for this if else, try to either use match or isolate your happy case

lib/cli/tests.rs
```diff
@@ -110,7 +126,8 @@ fn should_create_deploy() {
     let mut output = Vec::new();
 
     let deploy =
-        deploy::with_payment_and_session(deploy_params, payment_params, session_params).unwrap();
+        deploy::with_payment_and_session(deploy_params, payment_params, session_params, false)
```
gRoussac, line 129: Problem with flags like that is that they don't really say what they are just by reading here. So I would gladly see what they are instead of having to look for the definition so I would suggest to maybe create an explicit var even if it does look unnecessary.  
```
let allow_unsigned_deploy = false;
write deploy::with_payment_and_session(deploy_params, payment_params, session_params, allow_unsigned_deploy).unwrap();

File: src/deploy/make.rs [hunk 11]
```diff
@@ -27,14 +27,14 @@ impl ClientCommand for MakeDeploy {
             .display_order(display_order);
         let subcommand = creation_common::apply_common_session_options(subcommand);
         let subcommand = creation_common::apply_common_payment_options(subcommand, None);
-        creation_common::apply_common_creation_options(subcommand, false)
+        creation_common::apply_common_creation_options(subcommand, false, false)
```
gRoussac, line 30: Same comment here it's bit hard to guess what are those booleans until you go to definition, so I would creat vars here to be more explicit. Matter of taste but .required(false) is far more explicit to read.

require_secret_key: bool,
include_node_address: bool,

src/deploy/sign.rs
```diff
@@ -33,7 +33,7 @@ impl ClientCommand for SignDeploy {
 
     async fn run(matches: &ArgMatches) -> Result<Success, CliError> {
         let input_path = creation_common::input::get(matches);
-        let secret_key = common::secret_key::get(matches);
+        let secret_key = common::secret_key::get(matches).unwrap_or_default();
```
gRoussac, line 36: Not sure what is default on secret key so I would rather manage the error or be explicit what is the default with

Fraser999 (approved):

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "casper-client"
 version = "2.0.0" # when updating, also update 'html_root_url' in lib.rs
-authors = ["Marc Brinkmann <marc@casperlabs.io>", "Fraser Hutchison <fraser@casperlabs.io>"]
+authors = ["Marc Brinkmann <marc@casperlabs.io>", "Fraser Hutchison <fraser@casperlabs.io>",]
```
bpr, line 4: You can add yourself to this list Zach. I'd like you to be the "owner" of the client going forward.

Cargo.toml
```diff
@@ -29,12 +29,12 @@ clap = { version = "4", features = ["cargo", "deprecated", "wrap_help"] }
...
 derp = "0.0.14"
 ed25519-dalek = { version = "1", default-features = false }
-getrandom = "0.2.8"
+getrandom = "0.2.10"
 hex-buffer-serde = "0.4.0"
 humantime = "2"
 itertools = "0.10.5"
 jsonrpc-lite = "0.6.0"
-k256 = { version = "0.7.3", features = ["arithmetic", "ecdsa", "sha256", "zeroize"] }
+k256 = { version = "0.7.3", features = ["arithmetic", "ecdsa", "sha256", "zeroize", ] }
```
bpr, line 37: Are a lot of these from a run of `cargo fmt`?

bpr (approved):
LGTM

PR #93, opened 2023-08-08
Transfers - extra session args

### LINKED ISSUE #67: Support extra session args in client transfer subcommands
The client needs to support --session-arg in the transfer and make-transfer subcommands.

Addressing the following request:

How to specify targetAccountHex in transfer shown in this deploy?
https://cspr.live/deploy/741ac7b56fef49d619aa1459c74d13c3081c2f38d41139e30ba955f1d92e675a
I didn't find argument for casper-client transfer . Is it from session transfer? if so how to specify this argument?

copied from: https://github.com/casper-network/casper-node/issues/1949

hoffmannjan:
For now it only supports `--session-arg` like mentioned in a [ticket](https://github.com/casper-ecosystem/casper-client-rs/issues/67). 

I would consider supporting also JSON or complex args but was not sure about this - that's why for now I made `parse::arg_simple::get` public and use it in `new_transfer` method. If we would like to support all of the types then I can add different layer of abstraction dedicated to extending `Transfer` args.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.  The format
 ## Unreleased
 
 ### Added
+* Support passing `--session-arg` into `make-transfer` and `transfer` subcommands.
 * Add support for crafting unsigned deploys and transfers by providing an account, but not seccret key, to the `make-deploy` and `make-transfer` subcommands.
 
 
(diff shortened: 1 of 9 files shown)


hoffmannjan:
After discussion with @Fraser999 - we want to add support also for JSONs args - so I will add a new parser that supports both JSON and simple-args rather than calling manually `parse::arg_simple::get`.

Also we want to get rid of args-complex - listed [here](https://github.com/casper-ecosystem/casper-client-rs/issues/94)

Issue #55, opened 2023-03-23
Support providing account hashes as an alternative to public keys throughout the casper-client where this makes sense


RitaMAllenCA:
Some of the subcommands (e.g. `get-account`) which currently only support providing a public key should be enhanced to support providing an account hash instead.  For reference, see how this is handled in `query-balance`.

bpr:
There needs to be 

1. A change in casper node ./node/src/components/rpc_server/rpcs/state.rs, where GetAccountInfoParams is defined, to handle the logic of a PurseIdentifier instead of a public key. 

2. Doc and maybe other changeto support a modified client api, and JSON rpc call, if we're changing state_get_acount_info, which backs get-account.

I've already made all of the changes in client, which is in review.

RitaMAllenCA:
I've talked with Fraser about the ticket and we've discovered that changes are still needed on the node to update the RPC handler for state_get_account_info.
I just wanted to check where this work would be in the pipeline for the core team.
Additionally, If it makes sense for the timelines on the core team, Fraser and I discussed potentially pairing to get those changes made. Rita, if Fraser can spare some cycles and you approve, I would love the opportunity!
@RBYGLLS

PR #99, opened 2023-08-17
Account identifier

zacshowa:
# Summary of work done:
- Update the get-account command to accept an account identifier.
- Add the `AccountIdentifier` enum to support this update.
- Ensure `AccountIdentifier` has attributes to allow serde to properly format it and avoid breaking changes to the RPC API

### Sample requests:

Request for an account using a public key from an NCTL network
```
{
  "jsonrpc": "2.0",
  "method": "state_get_account_info",
  "params": {
    "public_key": "01fbabf372fe762ad840834f7dab5d88e522ffa39c9f53d1e991056cef110faeb2",
    "block_identifier": null
  },
  "id": -2271248039839344330
}
```

Associated output:

```
{
  "jsonrpc": "2.0",
  "result": {
    "api_version": "1.0.0",
    "account": {
      "account_hash": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
      "named_keys": [],
      "main_purse": "uref-cf8a03704e4865323fd1b02166e52302ce8fc4c931d76d04e6faea33ee52d93e-007",
      "associated_keys": [
        {
          "account_hash": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
          "weight": 1
        }
      ],
      "action_thresholds": {
        "deployment": 1,
        "key_management": 1
      }
    },
    "merkle_proof": "[2228 hex chars]"
  },
  "id": -2271248039839344330
}
```

Request for the same account using an account hash instead:

```
{
  "jsonrpc": "2.0",
  "method": "state_get_account_info",
  "params": {
    "public_key": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
    "block_identifier": null
  },
  "id": -3436401869523751496
}
```

Associated output:

```
{
  "jsonrpc": "2.0",
  "result": {
    "api_version": "1.0.0",
    "account": {
      "account_hash": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
      "named_keys": [],
      "main_purse": "uref-cf8a03704e4865323fd1b02166e52302ce8fc4c931d76d04e6faea33ee52d93e-007",
      "associated_keys": [
        {
          "account_hash": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
          "weight": 1
        }
      ],
      "action_thresholds": {
        "deployment": 1,
        "key_management": 1
      }
    },
    "merkle_proof": "[2228 hex chars]"
  },
  "id": -3436401869523751496
}
```

# Associated ticket(s):
- #55  
- [Support providing account hashes as an alternative to public keys throughout the casper-client where this makes sense](https://app.zenhub.com/workspaces/rust-sdk-64b185bf2cb87924e7759d9d/issues/gh/casper-ecosystem/casper-client-rs/55)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ async-trait = "0.1.59"
 base16 = "0.2.1"
 base64 = "0.13.1"
 casper-hashing = "2.0.0"
-casper-types = { version = "3.0.0", features = ["std"] }
+casper-types = { version = "3.0.0", features = ["std", "json-schema"] }
 clap = { version = "4", features = ["cargo", "deprecated", "wrap_help"] }
 clap_complete = "4"
 derp = "0.0.14"
@@ -55,7 +55,7 @@ vergen = { version = "7", default-features = false, features = ["git"] }
 
 [patch.crates-io]
 casper-hashing = { git = "https://github.com/casper-network/casper-node", branch = "release-1.5.0-rc.1" }
-casper-types = { git = "https://github.com/casper-network/casper-node", branch = "release-1.5.0-rc.1" }
+casper-types = { git = "https://github.com/casper-network/casper-node", branch = "release-1.5.0-rc.1"}
 
 [package.metadata.deb]
 features = ["vendored-openssl"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "casper-client"
 version = "2.0.0" # when updating, also update 'html_root_url' in lib.rs
-authors = ["Marc Brinkmann <marc@casperlabs.io>", "Fraser Hutchison <fraser@casperlabs.io>",]
+authors = ["Marc Brinkmann <marc@casperlabs.io>", "Fraser Hutchison <fraser@casperlabs.io>", "Zachary Showalter <zach@casperlabs.io>"]
 edition = "2021"
 description = "A client library and binary for interacting with the Casper network"
 documentation = "https://docs.rs/casper-client"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,40 +22,34 @@ doc = false
 [dependencies]
 async-trait = "0.1.59"
 base16 = "0.2.1"
-base64 = "0.13.1"
 casper-hashing = "2.0.0"
-casper-types = { version = "3.0.0", features = ["std", "json-schema"] }
+casper-types = { version = "3.0.0", features = ["std"] }
 clap = { version = "4", features = ["cargo", "deprecated", "wrap_help"] }
 clap_complete = "4"
-derp = "0.0.14"
-ed25519-dalek = { version = "1", default-features = false }
-getrandom = "0.2.10"
 hex-buffer-serde = "0.4.0"
 humantime = "2"
-itertools = "0.10.5"
+itertools = "0.11.0"
 jsonrpc-lite = "0.6.0"
-k256 = { version = "0.7.3", features = ["arithmetic", "ecdsa", "sha256", "zeroize", ] }
 num-traits = "0.2.15"
 once_cell = "1"
-pem = "1"
 rand = "0.8.5"
 reqwest = { version = "0.11.13", features = ["json"] }
 schemars = "0.8"
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde_json = { version = "1", features = ["preserve_order"] }
-signature = "1"
-tempfile = "3"
-thiserror = "=1.0.34"
+thiserror = "1.0.34"
 tokio = { version = "1.23.0", features = ["macros", "net", "rt-multi-thread", "sync", "time", ]}
 uint = "0.9.4"
-untrusted = "0.9.0"
+
+[dev-dependencies]
+tempfile = "3.7.1"
 
 [build-dependencies]
 vergen = { version = "7", default-features = false, features = ["git"] }
 
 [patch.crates-io]
-casper-hashing = { git = "https://github.com/casper-network/casper-node", branch = "release-1.5.0-rc.1" }
-casper-types = { git = "https://github.com/casper-network/casper-node", branch = "release-1.5.0-rc.1"}
+casper-hashing = { git = "https://github.com/casper-network/casper-node", branch = "dev" }
+casper-types = { git = "https://github.com/casper-network/casper-node", branch = "dev"}
 
 [package.metadata.deb]
 features = ["vendored-openssl"]
--- a/lib/cli/parse.rs
+++ b/lib/cli/parse.rs
@@ -15,7 +15,8 @@ use casper_types::{
 use super::{simple_args, CliError, PaymentStrParams, SessionStrParams};
 use crate::{
     types::{BlockHash, DeployHash, ExecutableDeployItem, TimeDiff, Timestamp},
-    BlockIdentifier, GlobalStateIdentifier, JsonRpcId, OutputKind, PurseIdentifier, Verbosity, AccountIdentifier
+    AccountIdentifier, BlockIdentifier, GlobalStateIdentifier, JsonRpcId, OutputKind,
+    PurseIdentifier, Verbosity,
 };
 
 pub(super) fn rpc_id(maybe_rpc_id: &str) -> JsonRpcId {
@@ -776,26 +777,29 @@ pub(super) fn account_identifier(account_identifier: &str) -> Result<AccountIden
     const ACCOUNT_HASH_PREFIX: &str = "account-hash-";
 
     if account_identifier.is_empty() {
-           return Err(CliError::InvalidArgument {
+        return Err(CliError::InvalidArgument {
             context: "account_identifier",
-            error: "cannot be empty string".to_string()});
+            error: "cannot be empty string".to_string(),
+        });
     }
 
     if account_identifier.starts_with(ACCOUNT_HASH_PREFIX) {
-        let account_hash = AccountHash::from_formatted_str(account_identifier).map_err(|error| {
-            CliError::FailedToParseAccountHash {
-                context: "account_identifier",
-                error,
-            }
-        })?;
+        let account_hash =
+            AccountHash::from_formatted_str(account_identifier).map_err(|error| {
+                CliError::FailedToParseAccountHash {
+                    context: "account_identifier",
+                    error,
+                }
+            })?;
         return Ok(AccountIdentifier::AccountHash(account_hash));
     }
 
-    let public_key =
-        PublicKey::from_hex(account_identifier).map_err(|error| CliError::FailedToParsePublicKey {
+    let public_key = PublicKey::from_hex(account_identifier).map_err(|error| {
+        CliError::FailedToParsePublicKey {
             context: "account_identifier".to_string(),
             error,
-        })?;
+        }
+    })?;
     Ok(AccountIdentifier::PublicKey(public_key))
 }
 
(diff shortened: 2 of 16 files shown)


lib/rpcs/v1_4_5/get_account.rs
```diff
@@ -1,22 +1,39 @@
...
+    PublicKey(PublicKey),
+    /// The account hash of an account
+    AccountHash(AccountHash),
+}
+
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
 pub(crate) struct GetAccountParams {
-    public_key: PublicKey,
+    ///The identifier of an Account. (named public key to match the JSON-RPC API)
```
gRoussac, line 12: Hi, it seems you switch from something like public to account mostly because you are adding AccountHash. An then the api is not following so personally I would have made `pub enum PublicKeyIdentifier` because behind a key or an account hash are identifiers of a public (key whatever its form) identifier. Account is to me something else that derives from the public key or account hash but naming things is always very subjective.

and so 
```rust
 pub(crate) fn new(
        public_key: PublicKeyIdentifier,
        
        ....
        
         GetAccountParams {
            public_key,
            block_identifier,
        }`

File: lib/cli/parse.rs [hunk 2]
```diff
@@ -771,6 +772,37 @@ pub(super) fn purse_identifier(purse_id: &str) -> Result<PurseIdentifier, CliErr
     Ok(PurseIdentifier::MainPurseUnderPublicKey(public_key))
 }
 
+/// Add
```
gRoussac: This comment is strange. I am looking were tests are affected by this change on get_info as it seems to me get_info only receives old public key so is parse::account_identifier(account_identifier) well tested with an account hash ?

gRoussac (approved):
Except the naming little conflict of names with the rpc lgtm

lib/rpcs/v1_4_5/get_account.rs
gRoussac: or maybe 

```rust

 pub(crate) fn new(
        public_key_identifier: PublicKeyIdentifier,
        
        ....
        
         GetAccountParams {
            public_key_identifier,,
            block_identifier,
        }`

```
because this does not read very well

``` 
 "params": {
    "public_key": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
    "block_identifier": null
  },
  ```
  But if you can break the API
  
  ``` 
 "params": {
    "public_key_identifier": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
    "block_identifier": null
  },
  
  ```
  
  or indeed account_identifier
  
```
 "params": {
    "account_identifier": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
    "block_identifier": null
  },

lib/cli/parse.rs
zacshowa: I'll add unit tests here for this method. It's well tested manually but not in the code.

lib/rpcs/v1_6_0.rs
```diff
@@ -0,0 +1,118 @@
+//! The JSON-RPC request and response types at v1.5.0 of casper-node.
```
zacshowa, line 1: For any readers of this PR, this is just re-exporting the previous RPC functions, and additionally exposing the new AccountIdentifier for version 1.6.0 of the node.

lib/rpcs/v1_4_5/get_account.rs
zacshowa: The reason for naming it `AccountIdentifier` is that a public key is, in the nodes eyes, really an identifier of an account on the network.

That is to say an account has a set of identifiers that includes an account hash, and a public key/private key pair.

Calling it a `PublicKeyIdentifier` seems inaccurate as the entity we are asking the node to retrieve isn't a public key, rather it's all of the details related to an account associated with either a public key or an account hash.

gRoussac: To me an account has a public ( any key form underneath) identifier that can take two forms right now, a public key or account hash.
Anyway it seems to me introducing the line `public_key: account_identifier,` is somehow a consequence of having to modify an arg and not adding an arg "account_identifier" to the structs and somehow deprecate "public_key" arg and also the api answer should also be answering the same way.

```
    "account": {
      "account_identifier": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e", // or 0x key
```

or both

```
      "account_hash": "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e",
      "public_key": "0x...",

File: Cargo.toml [hunk 4]
```diff
@@ -24,7 +24,7 @@ async-trait = "0.1.59"
 base16 = "0.2.1"
 base64 = "0.13.1"
 casper-hashing = "2.0.0"
-casper-types = { version = "3.0.0", features = ["std"] }
+casper-types = { version = "3.0.0", features = ["std", "json-schema"] }
```
Fraser999: It shouldn't hurt, but this addition also shouldn't be needed I think.

lib/cli/parse.rs
```diff
@@ -1492,4 +1526,38 @@ mod tests {
             }
         }
     }
+    mod account_identifier{
```
Fraser999: Nit: we'd normally have a blank line before a new module like this, and also before each test case.

lib/cli/parse.rs
```diff
@@ -1492,4 +1526,38 @@ mod tests {
...
+        use casper_types::account::AccountHash;
+        use crate::rpcs::v1_6_0::get_account::AccountIdentifier;
```
Fraser999: The imports aren't required given the `use super::*;` below.

lib/cli/parse.rs
```diff
@@ -1492,4 +1526,38 @@ mod tests {
...
+        use super::*;
+        #[test]
+        pub fn should_parse_valid_account_hash() {
+            let account_hash = "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e";
+            let parsed = account_identifier(account_hash).unwrap();
+            let _expected = AccountHash::from_formatted_str(account_hash).unwrap();
```
Fraser999: Nit: the leading underscore is normally only applied to unused variables. (Same comment on test below).

lib/lib.rs
```diff
@@ -72,7 +72,8 @@ use rpcs::{
         GetStateRootHashResult, GetValidatorChangesResult, ListRpcsResult, PutDeployResult,
         QueryBalanceResult, QueryGlobalStateResult, SpeculativeExecResult,
     },
-    v1_5_0::{
+    v1_6_0::get_account::AccountIdentifier,
```
Fraser999: Nit: this can be merged into the existing `v1_6_0::get_account` imports just below here.

lib/rpcs/v1_4_5/get_account.rs
```diff
@@ -1,22 +1,39 @@
+use schemars::JsonSchema;
```
Fraser999: This shouldn't be required.  We could look to remove it from any other types we apply it to in a follow-up PR.

lib/rpcs/v1_4_5/get_account.rs
Fraser999: If we apply the serde alias as you suggest [here](https://github.com/casper-network/casper-node/pull/4218#discussion_r1297674858) @zacshowa, we should be able to define this field in the client struct as `account_identifier: AccountIdentifier` without even applying the alias.

lib/rpcs/v1_4_5/get_account.rs
Fraser999, line 1: This file should remain unchanged so it remains fully compatible with a 1.4.5 node.  We should instead have a new folder `lib/rpcs/v1_6_0` containing what is now this file after your changes.

src/get_account.rs
```diff
@@ -41,9 +41,9 @@ impl ClientCommand for GetAccount {
                 DisplayOrder::BlockIdentifier as usize,
                 true,
             ))
-            .arg(common::public_key::arg(
-                DisplayOrder::PublicKey as usize,
-                PUBLIC_KEY_IS_REQUIRED,
+            .arg(common::account_identifier::arg(
+                DisplayOrder::AccountIdentifier as usize,
+                ACCOUNT_IDENTIFIER_IS_REQUIRED,
             ))
```
Fraser999: To retain backwards compatibility at the CLI level:
```suggestion
            .arg(
                common::account_identifier::arg(
                    DisplayOrder::AccountIdentifier as usize,
                    ACCOUNT_IDENTIFIER_IS_REQUIRED,
                )
                .alias(ACCOUNT_IDENTIFIER_ALIAS)
                .short_alias(ACCOUNT_IDENTIFIER_SHORT_ALIAS),
            )
```
where we also have
```rust
const ACCOUNT_IDENTIFIER_ALIAS: &str = "public-key";
const ACCOUNT_IDENTIFIER_SHORT_ALIAS: char = 'p';
```

Fraser999 (commented):
We should update `lib/rpcs/results.rs` to use the new `v1_6_0` types everywhere.

Looking good otherwise - mostly nitpicks.

lib/rpcs/v1_6_0/get_account.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// The public key of an account
+    PublicKey(PublicKey),
+    /// The account hash of an account
+    AccountHash(AccountHash),
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct GetAccountParams {
+    ///The identifier of an Account. (named public key to match the JSON-RPC API)
```
Fraser999:
```suggestion
    /// The identifier of an Account.
```

lib/cli/parse.rs
```diff
@@ -1492,4 +1526,45 @@ mod tests {
...
+        #[test]
+        pub fn should_parse_valid_account_hash() {
+            let account_hash =
+                "account-hash-c029c14904b870e64c1d443d428c606740e82f341bea0f8542ca6494cef1383e";
+            let parsed = account_identifier(account_hash).unwrap();
+            //This lint should be re-enabled once this bug is fixed. Currently, clippy erroneously flags this variable as unused on both lines.
+            #[allow(unused_variables)]
+            let expected = AccountHash::from_formatted_str(account_hash).unwrap();
+            //The _ should be removed from this statement once the above clippy bug is fixed.
+            assert!(matches!(parsed, AccountIdentifier::AccountHash(_expected)));
```
Fraser999: An unused warning from inside a `matches!` macro is normally a really strong hint that the macro isn't doing what you think it is! :)  Just now, we're only checking that `parsed` has the correct variant - that it matches `AccountIdentifier::AccountHash`, but we're not checking the actual account hash's value.

In this case, I think the best approach is to derive `PartialEq` and `Eq` for `AccountIdentifier`, and change this test to look for equality rather than just matching on the `AccountIdentifier` variant.  So, e.g.
```rust
let expected = AccountIdentifier::AccountHash(
    AccountHash::from_formatted_str(account_hash).unwrap(),
);
assert_eq!(parsed, expected);
```

Same comment for the test below.

Fraser999 (commented):
Two final minor issues and then we're gtg.

Fraser999 (approved):
LGTM

PR #101, opened 2023-09-01
Merge dev

Fraser999:
This brings `dev` into `feat-track-node-2.0` in order to fix a build breakage due to an update to the `casper-types` dep specified in `[patch.crates-io]`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ async-trait = "0.1.59"
 base16 = "0.2.1"
 base64 = "0.13.1"
 casper-hashing = "2.0.0"
-casper-types = { version = "3.0.0", features = ["std"] }
+casper-types = { version = "3.0.0", features = ["std", "json-schema"] }
 clap = { version = "4", features = ["cargo", "deprecated", "wrap_help"] }
 clap_complete = "4"
 derp = "0.0.14"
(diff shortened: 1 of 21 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "casper-client"
 version = "2.0.0" # when updating, also update 'html_root_url' in lib.rs
-authors = ["Marc Brinkmann <marc@casperlabs.io>", "Fraser Hutchison <fraser@casperlabs.io>",]
+authors = ["Marc Brinkmann <marc@casperlabs.io>", "Fraser Hutchison <fraser@casperlabs.io>", "Zachary Showalter <zach@casperlabs.io>"]
```
bpr, line 4: Excellent, thanks Fraser!

zacshowa (approved):
LGTM!

lib/cli/parse.rs
```diff
@@ -376,7 +377,7 @@ macro_rules! check_exactly_one_not_empty {
                 });
             }
         } else {
-            let mut non_empty_fields_with_values = vec![$((stringify!($x), $x)),+]
+            let mut non_empty_fields_with_values = [$((stringify!($x), $x)),+]
```
bpr, line 382: Gotta admit,  I had to look this up. Not just the change, but the original code. So, this is subtle enough to deserve either rewriting into simpler code, or a comment.

I'd say the same about a complex regexp too. Any place where the average reader would do a double take deserves a comment.

bpr (approved):
Please add a comment where I asked, explaining the unusual (IMO) code. Other than that, LGTM

lib/cli/parse.rs
Fraser999: Done in [157832d](https://github.com/casper-ecosystem/casper-client-rs/pull/101/commits/157832dd2d6f4e4b85b952e8d456388064ec0206).

PR #106, opened 2023-09-16
Add sdk feature into feat 2.0

gRoussac:
Right now this feature only build with --lib

Mainly adds some `#[cfg(feature = "sdk")]` where it is possible or not to use io/fs, or prevent sending to std with `json_pretty_print` and returning deploys 
Removes rt-multi-thread, `The default runtime flavor is 'multi_thread', but the 'rt-multi-thread' feature is disabled.` switching (flavor = "current_thread") to tokio main

<s>I could not make it compile without this change in cargo.toml related to casper types
`getrandom = { version = "0.2.10", features = ["js"] }` getrandom probably needs to be replaced by something else.</s>
Removing `[patch.crates-io]` as I believe this bug was fixed and this is not needed anymore.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ k256 = { version = "0.7.3", features = [
     "zeroize",
 ] }
 num-traits = "0.2.15"
-once_cell = "1"
+once_cell = { version = "1" }
 pem = "1"
 rand = "0.8.5"
 reqwest = { version = "0.11.13", features = ["json"] }
@@ -65,7 +65,7 @@ untrusted = "0.9.0"
 
 [features]
 default = ["tokio", "clap", "casper-types/std", "casper-types/json-schema"]
-sdk = ["casper-types/json-schema"]
+sdk = []
 
 [build-dependencies]
 vergen = { version = "7", default-features = false, features = ["git"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,11 +27,10 @@ doc = false
 async-trait = "0.1.59"
 base16 = "0.2.1"
 base64 = "0.13.1"
-casper-types = { version = "3.0.0", features = ["std", "json-schema"] }
+casper-types = { version = "3.0.0", default-features = false }
 clap = { version = "4", features = ["cargo", "deprecated"], optional = true }
-clap_complete = "4"
+clap_complete = { version = "4" }
 derp = "0.0.14"
-ed25519-dalek = { version = "1", default-features = false }
 getrandom = { version = "0.2.10", features = ["js"] }
 hex-buffer-serde = "0.4.0"
 humantime = "2"
@@ -48,7 +47,7 @@ once_cell = "1"
 pem = "1"
 rand = "0.8.5"
 reqwest = { version = "0.11.13", features = ["json"] }
-schemars = "0.8"
+schemars = { version = "0.8" }
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde-map-to-array = "1.1.0"
 serde_json = { version = "1", features = ["preserve_order"] }
@@ -65,9 +64,8 @@ uint = "0.9.4"
 untrusted = "0.9.0"
 
 [features]
-default = ["casper-types", "tokio", "clap"]
-casper-types = ["casper-types/std"]
-sdk = []
+default = ["tokio", "clap", "casper-types/std", "casper-types/json-schema"]
+sdk = ["casper-types/json-schema"]
 
 [build-dependencies]
 vergen = { version = "7", default-features = false, features = ["git"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ num-traits = "0.2.15"
 once_cell = "1"
 rand = "0.8.5"
 reqwest = { version = "0.11.13", features = ["json"] }
-schemars = "0.8"
+schemars = "0.8.13"
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde-map-to-array = "1.1.0"
 serde_json = { version = "1", features = ["preserve_order"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,8 +57,7 @@ tempfile = "3"
 thiserror = "=1.0.34"
 tokio = { version = "1.23.0", features = [
     "macros",
-    "net",
-    "rt-multi-thread",
+    "rt",
     "sync",
     "time",
 ], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ doc = false
 [dependencies]
 async-trait = "0.1.59"
 base16 = "0.2.1"
-casper-types = { version = "3.0.0", git = "https://github.com/casper-network/casper-node.git", branch = "rustSDK", features = [
+casper-types = { version = "3.0.0", git = "https://github.com/casper-network/casper-node.git", branch = "feat-2.0", features = [
     "std",
     "json-schema",
 ] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ async-trait = "0.1.59"
 base16 = "0.2.1"
 base64 = "0.13.1"
 casper-types = { version = "3.0.0", git = "https://github.com/casper-network/casper-node.git", branch = "rustSDK", default-features = false, features = [
+    "sdk",
     "std",
     "json-schema",
 ] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ doc = false
 async-trait = "0.1.59"
 base16 = "0.2.1"
 base64 = "0.13.1"
-casper-types = { version = "3.0.0", git = "https://github.com/casper-network/casper-node.git", branch = "rustSDK", default-features = false, features = [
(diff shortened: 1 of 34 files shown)


bpr (approved):
LGTM, but please wait for @Fraser999 and @zacshowa to review before merging. There's a lot of changes and I think we should tread carefully here.

lib/cli/deploy.rs
```diff
@@ -123,3 +98,38 @@ pub fn new_transfer(
         .map_err(crate::Error::from)?;
     Ok(deploy)
 }
+
+fn get_maybe_secret_key(
```
zacshowa, line 103: The change to this function is interesting.

Perhaps we could look to overload the function by conditionally compiling 1 of 2 functions of the same name behind feature flags as opposed to modifying the body of one function for readability.

lib/cli/error.rs
```diff
@@ -112,7 +112,7 @@ pub enum CliError {
     ConflictingArguments {
         /// Contextual description of where this error occurred including relevant paths,
         /// filenames, etc.
-        context: &'static str,
+        context: String,
```
zacshowa, line 115: I don't think we need a `String` here. Is there a reason you need this to change in the clients lib instead of doing any needed conversion in SDK land?

lib/cli/parse.rs
```diff
@@ -1,22 +1,26 @@
...
+    account::AccountHash,
+    bytesrepr::{self, Bytes},
+    crypto, AsymmetricType, BlockHash, CLValue, DeployHash, Digest, ExecutableDeployItem, HashAddr,
+    Key, NamedArg, PublicKey, RuntimeArgs, TimeDiff, Timestamp, UIntParseError, URef, U512,
+};
+use rand::Rng;
+use serde::{self, Deserialize};
+#[cfg(not(any(feature = "sdk")))]
+use std::path::Path;
+use std::{convert::TryInto, fs, io, str::FromStr};
```
zacshowa: I would prefer that imports stay in the order of 

```
std

external crates

crates originating from Casper

internal crate imports
```

lib/cli.rs
```diff
@@ -30,14 +30,21 @@ mod parse;
 mod payment_str_params;
 mod session_str_params;
 mod simple_args;
+#[cfg(feature = "sdk")]
```
zacshowa, line 33: I worry that making these public in the client, rather than re-implementing them in the SDK could create external dependencies on this code. This would cause any changes to these items to be breaking and cause it to be a headache for future maintenance of the client. Perhaps these items can just be copied over to the SDK itself.

src/main.rs
```diff
@@ -142,7 +142,7 @@ fn cli() -> Command {
         ))
 }
 
-#[tokio::main]
+#[tokio::main(flavor = "current_thread")]
```
zacshowa, line 145: Is there a reason we need to make the tokio runtime for the client single threaded?

Cargo.toml
```diff
@@ -22,9 +27,13 @@ doc = false
 [dependencies]
 async-trait = "0.1.59"
 base16 = "0.2.1"
-casper-types = { version = "3.0.0", features = ["std"] }
-clap = { version = "4", features = ["cargo", "deprecated", "wrap_help"] }
+casper-types = { version = "3.0.0", git = "https://github.com/casper-network/casper-node.git", branch = "feat-2.0", features = [
```
zacshowa, line 30: I don't have a strong preference on how the cargo.toml is formatted. I'm just curious what lad to the changes

zacshowa (changes requested):
A few things could use some changes/clarification.

lib/cli/error.rs
gRoussac: Yes I modified the Error to receive a computed string. The error is dynamic, some other errors have context String too. https://github.com/casper-ecosystem/casper-client-rs/pull/106/files#diff-657c4a5be21b7e27f49860ca9e5ce63f061285609263aad05c184c1af2e12fa9R326

lib/cli/parse.rs
gRoussac: My VsCode is reorganizing imports on save, I am not sure about the rule applied, I guess it's alphabetic. I can try to manually re assort them.

lib/cli.rs
gRoussac: I didn't re implement methods, every rpc call, deploys everything is relying on the casper client. The purpose is to have the client served as a library, not to duplicate code to have double maintenance, already few methods are exposed, I don't see why 2 more would be problematic. If the casper client/node evolve, then the sdk is following.

src/main.rs
gRoussac: Fraser and I cleaned up imports of tokio, removing *un*necessary mutli thread rt-multi-thread to rt and the compiler asked to add that attribute.  

```
   Compiling casper-client v2.0.0 (/casper/casper-client-rs)
error: The default runtime flavor is `multi_thread`, but the `rt-multi-thread` feature is disabled.
   --> src/main.rs:145:1
    |
145 | #[tokio::main]
    | ^^^^^^^^^^^^^^
    |
    = note: this error originates in the attribute macro `tokio::main` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `casper-client` (bin "casper-client") due to previous error
```

Cargo.toml
gRoussac: I needed to compile with std and then to wasm so I needed to remove calls to io/fs and system calls. So with Fraser we tried to see what were the smallest changes. It was mostly clap and schemars giving issues. Then the rest is adding the feature it self (which is a restricting feature while a feature should actually be additive) and libs `"cdylib", "rlib"`.
Biggest change is `features = ["js"]  to getrandom`, I am not sure how to workaround that.

```shell

   Compiling getrandom v0.2.10
error: the wasm*-unknown-unknown targets are not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /opt/rust/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.10/src/lib.rs:285:9
    |
285 | /         compile_error!("the wasm*-unknown-unknown targets are not supported by \
286 | |                         default, you may need to enable the \"js\" feature. \
287 | |                         For more information see: \
288 | |                         https://docs.rs/getrandom/#webassembly-support");
    | |________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /opt/rust/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.10/src/lib.rs:341:9
    |
341 |         imp::getrandom_inner(dest)?;
    |         ^^^ use of undeclared crate or module `imp`
```
    
The formatting changes are my VsCode with probably Better Toml extention that is making formatting on save.

lib/cli/deploy.rs
gRoussac: Well mostly this code was repeated two times in the file and we already discussed this condition 

```
 let maybe_secret_key = if allow_unsigned_deploy && deploy_params.secret_key.is_empty() {
        None
    } else if deploy_params.secret_key.is_empty() && !allow_unsigned_deploy {
   
 ````
 
 is imo more idiomatic in if A / else if B / else. It's quite complicate to have cargo check and cargo clippy accepting overloads, I tried to have 1/2 methods with both config accordingly and that never passed cargo check/clippy. Most of times the only accepted way is to simply have #[cfg] working like if statements. Otherwise most of the times a function will be missing on compilation if you have like --features-all

Cargo.toml
gRoussac: I removed "json-schema" that was added back apparently.

gRoussac: I included `getrandom/js` only for sdk feature, this should be less invasive for default build.

gRoussac: Actually `getrandom` was removed from dev too completely already I missed that so this issue does not exist anymore.

bpr (approved):
LGTM

lib/cli/parse.rs
gRoussac: I updated the file please check it fits the reco 

https://github.com/casper-ecosystem/casper-client-rs/pull/106/commits/834032ca5646080e3d62655bd77656e666725db7

src/main.rs
gRoussac: @Fraser999 Can you please confirm we want to get rid of rt-multi-thread in the client ? thanks

lib/cli/deploy.rs
zacshowa: The structure of the if else is fine, It's functionally equivalent and should work fine.
I do think that if you like the style suggestion I put on the sister PR against the node I would love to see the same style applied to the feature flag here. If you don't want to go with that style, consider the PR approved.

zacshowa (commented):
Like I said in the comment on deploy if you don't like the style suggestion consider the PR approved.

zacshowa (commented):
Just trying to remove the change request now.

zacshowa (approved):

lib/cli/deploy.rs
gRoussac: Casper client do not seem to have cargo clippy so I can duplicate this method in the client without side effects apparently.

Please check that commit https://github.com/casper-ecosystem/casper-client-rs/commit/4855e7ba3329eb38a960baaf2c62d5fa1121e387 thanks

PR #109, opened 2023-10-17
Proposition of changes into Casper Types to support a Rust SDK for 1.6

gRoussac:
This PR intends to pull changes made for the Rust SDK. It is adding a sdk feature and affects several methods or expose them accordingly to be able to compile the SDK.

This PR has a strong dependency on PR against node 1.6 https://github.com/casper-network/casper-node/pull/4348

Right now this feature only build with --lib

Mainly adds some `#[cfg(feature = "sdk")]` where it is possible or not to use io/fs, or prevent sending to std with `json_pretty_print` and returning deploys 
Removes rt-multi-thread, `The default runtime flavor is 'multi_thread', but the 'rt-multi-thread' feature is disabled.` switching (flavor = "current_thread") to tokio main. Please check this change is legit.

 Had an issue with schemars, I had to fix the version to "=0.8.5" apparently with Casper node 1.6.
 
 For simplicity for now keeping new get_maybe_secret_key() as one function and not two as per previous PR for 2.0
 
 https://github.com/casper-ecosystem/rustSDK/issues/13

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,9 +40,9 @@ itertools = "0.11.0"
 jsonrpc-lite = "0.6.0"
 num-traits = "0.2.15"
 once_cell = "1"
-rand = "=0.8.5"
+rand = "0.8.5"
 reqwest = { version = "0.11.13", features = ["json"] }
-schemars = "0.8.5"
+schemars = "=0.8.5"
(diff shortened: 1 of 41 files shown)


gRoussac:
I need to reopen the PR from a new branch to keep the SDK compilation working

PR #110, opened 2023-10-17
Proposition of changes into Casper Types to support a Rust SDK for 1.6

gRoussac:
This PR intends to pull changes made for the Rust SDK. It is adding a `std-fs-io` default feature and affects several methods or expose them accordingly to be able to compile the client with 

````shell
cargo build --lib --no-default-features --target wasm32-unknown-unknown
````

This PR has a strong dependency on PR against node 1.6 https://github.com/casper-network/casper-node/pull/4348

Mainly adds some `#[cfg(feature = "std-fs-io")]` where it is possible or not to use io/fs, or prevent sending to std with `json_pretty_print` and returning deploys 
Removes rt-multi-thread, `The default runtime flavor is 'multi_thread', but the 'rt-multi-thread' feature is disabled.` switching (flavor = "current_thread") to tokio main. Please check this change is legit.
 
 For simplicity for now keeping new get_maybe_secret_key() as one function and not two as per previous PR for 2.0
 
 https://github.com/casper-ecosystem/rustSDK/issues/13

### DIFF
--- a/lib/types/deploy.rs
+++ b/lib/types/deploy.rs
@@ -406,9 +406,9 @@ impl<'a> DeployBuilder<'a> {
     ///     [`with_payment`](Self::with_payment)
     pub fn new<C: Into<String>>(chain_name: C, session: ExecutableDeployItem) -> Self {
         // Timestamp::now() not available with feature sdk, use default() instead
-        #[cfg(feature = "std-output")]
+        #[cfg(feature = "std-fs-io")]
         let timestamp = Timestamp::now();
-        #[cfg(not(feature = "std-output"))]
+        #[cfg(not(feature = "std-fs-io"))]
         let timestamp = Timestamp::default();
 
         DeployBuilder {
--- a/lib/types/deploy.rs
+++ b/lib/types/deploy.rs
@@ -196,11 +196,20 @@ impl ToBytes for Approval {
 #[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Serialize, Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
 pub struct Deploy {
-    hash: DeployHash,
-    header: DeployHeader,
-    payment: ExecutableDeployItem,
-    session: ExecutableDeployItem,
-    approvals: Vec<Approval>,
+    /// The unique identifier for the deploy.
+    pub hash: DeployHash,
+
+    /// The header information for the deploy.
+    pub header: DeployHeader,
+
+    /// The payment logic for the deploy.
+    pub payment: ExecutableDeployItem,
+
+    /// The session logic for the deploy.
+    pub session: ExecutableDeployItem,
+
+    /// List of approvals for the deploy.
+    pub approvals: Vec<Approval>,
 }
 
 /// Used when constructing a `Deploy`.
@@ -396,10 +405,15 @@ impl<'a> DeployBuilder<'a> {
     ///     [`with_standard_payment`](Self::with_standard_payment) or
     ///     [`with_payment`](Self::with_payment)
     pub fn new<C: Into<String>>(chain_name: C, session: ExecutableDeployItem) -> Self {
+        #[cfg(not(any(feature = "sdk")))]
+        let timestamp = Timestamp::now();
+        #[cfg(feature = "sdk")]
+        let timestamp = Timestamp::default();
+
         DeployBuilder {
             account: None,
             secret_key: None,
-            timestamp: Timestamp::now(),
+            timestamp,
             ttl: Deploy::DEFAULT_TTL,
             gas_price: Deploy::DEFAULT_GAS_PRICE,
             dependencies: vec![],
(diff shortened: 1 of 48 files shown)

### CODE REVIEW

**Review by zacshowa — APPROVED** [review]
I think, for the sake of expediency in getting this out for 1.6, these changes are fine. However, I want to push for scheduling work for 2.0 to make this an additive feature.

File: lib/types/deploy.rs [hunk 1]
```diff
@@ -390,10 +405,16 @@ impl<'a> DeployBuilder<'a> {
     ///     [`with_standard_payment`](Self::with_standard_payment) or
     ///     [`with_payment`](Self::with_payment)
     pub fn new<C: Into<String>>(chain_name: C, session: ExecutableDeployItem) -> Self {
+        // Timestamp::now() not available with feature sdk, use default() instead
+        #[cfg(feature = "std-fs-io")]
+        let timestamp = Timestamp::now();
+        #[cfg(not(feature = "std-fs-io"))]
+        let timestamp = Timestamp::default();
```
Fraser999: This is a problem we need to resolve.  Using `Timestamp::default();` will always yield an expired deploy.

Fraser999 (changes requested):
We need to resolve the issue mentioned in my comment.

lib/types/deploy.rs
gRoussac: I don't believe one would find a clock in non os environment but I could be wrong.
If you use the client as a lib without `std-fs-io` you will probably need a clock in your app to timestamp the deploy with the timestamp the user wants (or a default one), so time will come either from `chrono` crate in a rust app or from javascript `js-sys` in wasm-bindgen. So far now we have only the sdk using the client as lib both ways rust/wasm and so far that I remember anyway you will have to give a timestamp to the deploy builder anyway if your user wants a specific timestamp (and thus a default one by the way if he does not is not that harder).

the sdk implements that by mimicking the current client

````toml
wasm-bindgen = "*"
js-sys = "*"
chrono = "0.4"
````

````rust
    #[wasm_bindgen(js_name = "withTimestamp")]
    pub fn with_timestamp(&self, timestamp: &str, secret_key: Option<String>) -> Deploy {
        let mut timestamp = parse_timestamp(timestamp);
        if let Err(err) = &timestamp {
            error(&format!("Error parsing Timestamp: {}", err));
            timestamp = parse_timestamp(&get_current_timestamp(None));
        }
        self.build(BuildParams {
            secret_key,
            timestamp: Some(timestamp.unwrap()),
            ..Default::default()
        })
    }
```

with 

````rust
    fn build(&self, deploy_params: BuildParams) -> Deploy {
        let BuildParams {
            secret_key,
            chain_name,
            ttl,
            timestamp,
            session,
            payment,
            account,
        } = deploy_params;
        ....
        let mut deploy_builder = DeployBuilder::new(chain_name, session)
            .with_account(account.into())
            .with_payment(payment)
            .with_ttl(ttl)
            .with_timestamp(timestamp);
````

and as helper into the sdk you have

````rust
use chrono::{DateTime, NaiveDateTime, SecondsFormat, Utc};

/// Gets the current timestamp.
///
/// # Arguments
///
/// * `timestamp` - An optional timestamp value in milliseconds since the Unix epoch.
///
/// # Returns
///
/// A string containing the current timestamp in RFC3339 format.
pub fn get_current_timestamp(timestamp: Option<String>) -> String {
    let parsed_timestamp = timestamp.as_ref().and_then(|ts| ts.parse::<i64>().ok());
    let current_timestamp = parsed_timestamp
        .map(|parsed_time| {
            NaiveDateTime::from_timestamp_opt(parsed_time / 1000, 0)
                .map(|naive_time| DateTime::<Utc>::from_naive_utc_and_offset(naive_time, Utc))
                .unwrap_or_else(Utc::now)
        })
        .unwrap_or_else(Utc::now);
    current_timestamp.to_rfc3339_opts(SecondsFormat::Secs, true)
}
````

I assume that if you use the client that way, ie without std-io-fs, you know you will not have a clock from the OS and that timestamp might be 0 by default and this will not pass put_deploy.
I believe anyhow if you try to build deploys (outside the sdk which is the main purpose of the sdk) you will probably have to timestamp them as per demand from the user and not rely only on the default one so you will have to implement `with_timestamp` your self. And also the sdk "solves" this question by exposing a way to build deploy as a lib and also timestamp it or not, so in fine a dev should rather import the sdk into its rust app whether than directly the client it self s/he wants to build a deploy easily without having to re implement what the sdk does for the dev like deploy::build / with_timestamp / is_expired / is_valid etc .

Fraser999: I see your point.

I added a change to the deploy builder so that it errors in the case where the timestamp isn't provided rather than returning what would be a permanently invalid deploy.  The changes are in [9ff1da4](https://github.com/casper-ecosystem/casper-client-rs/pull/110/commits/9ff1da46207fd42c9d99e1621403598c9a12a63c).

Fraser999 (approved):
LGTM now.

Issue #120, opened 2023-11-14
Client Command Updates 1 of 3


RBYGLLS:
Add 3 new commands to the client using this infrastructure developed in the first ticket #119 (Parsing and Buidling Transactions in the client).

PR #131, opened 2023-12-08
make-transaction

zacshowa:
# Summary of work done:
- Add an initial make-transaction command that specifically targets creating transactions via a wasm file.

The command specifies the session account twice as I still need to identify and implement the proper inputs via clap arguments
Additionally, the command already supports outputting the transaction to a user specified file location, as long as the directory exists.
```
cargo run -- make-transaction --session-path /path/to/wasm/contract.wasm -p 100000000 --initiator-address 01cf97f65569cc9187ec8f4948dd608d36345fa3aee67694236d208ba458d28fa4 --session-account 01cf97f65569cc9187ec8f4948dd608d36345fa3aee67694236d208ba458d28fa4 --chain-name test

```

Associated output:

```
{
  "hash": "7c4e10ec490cb22c76cacdb92bd9cd9619c8ba90d42474640c8d4d5f8f02a53e",
  "header": {
    "chain_name": "test",
    "timestamp": "2023-12-08T00:38:54.615Z",
    "ttl": "30m",
    "body_hash": "80273309fbc5b0a00b20f45cbbaca648ef830509cd1e769b6c6a6a98714f4bae",
    "pricing_mode": {
      "GasPriceMultiplier": 1
    },
    "payment_amount": 100000000,
    "initiator_addr": {
      "PublicKey": "01cf97f65569cc9187ec8f4948dd608d36345fa3aee67694236d208ba458d28fa4"
    }
  },
  "body": {
    "args": [],
    "target": {
      "Session": {
        "kind": "Standard",
        "module_bytes": "Omitted in PR for the sake of readability",
        "runtime": "VmCasperV1"
      }
    },
    "entry_point": {
      "Custom": ""
    },
    "scheduling": "Standard"
  },
  "approvals": []
}

```

# Associated ticket(s):
- #120 
- [Client Command Updates 1 of 3](https://app.zenhub.com/workspaces/casperlabs-sdks-64b185bf2cb87924e7759d9d/issues/gh/casper-ecosystem/casper-client-rs/120)


src/transaction/creation_common.rs
```diff
@@ -0,0 +1,883 @@
...
+    SessionHash,
+    SessionName,
+    SessionPackageHash,
+    SessionPackageName,
+    SessionEntryPoint,
+    SessionVersion,
+    SessionTransfer,
+    SessionAccount,
+    PaymentAmmount,
+    PaymentMode,
```
Fraser999:
```suggestion
    PricingMode,
```

lib/cli.rs
```diff
@@ -294,6 +297,28 @@ pub fn make_transfer(
...
+/// and the file will not be written.
+pub fn make_transaction(
+    maybe_output_path: &str,
+    transaction_params: TransactionStrParams<'_>,
+    session_params: SessionStrParams<'_>,
+    payment_amount: &str,
+    force: bool,
+) -> Result<(), CliError>{
+    let output = parse::output_kind(maybe_output_path, force);
+    let deploy =
```
Fraser999:
```suggestion
    let txn =
```

lib/cli/transaction.rs
```diff
@@ -0,0 +1,53 @@
...
+        .with_chain_name(chain_name);
+
+    if let Some(secret_key) = &maybe_secret_key {
+        transaction_builder = transaction_builder.with_secret_key(secret_key);
+    }
+    if let Some(account) = maybe_session_account {
+        transaction_builder = transaction_builder.with_initiator_addr(InitiatorAddr::PublicKey(account));
+    }
+
+    let deploy = transaction_builder.build().map_err(crate::Error::from)?;
```
Fraser999:
```suggestion
    let txn = transaction_builder.build().map_err(crate::Error::from)?;
```

src/transaction/creation_common.rs
Fraser999, line 1: There are quite a few instances where the term "deploy" should be replaced by "transaction" in this file, particularly in the help strings.

src/transaction/creation_common.rs
```diff
@@ -0,0 +1,883 @@
...
+            .help(ARG_HELP)
+            .display_order(DisplayOrder::SessionCode as usize)
+    }
+
+    pub(in crate::transaction) fn get(matches: &ArgMatches) -> Option<&str> {
+        matches.get_one::<String>(ARG_NAME).map(String::as_str)
+    }
+}
+
+/// Handles providing the arg for and retrieval of simple session and payment args.
```
Fraser999:
```suggestion
/// Handles providing the arg for and retrieval of simple session args.
```

src/transaction/creation_common.rs
```diff
@@ -0,0 +1,883 @@
...
+            .long(name)
+            .required(false)
+            .action(ArgAction::Append)
+            .value_name(ARG_VALUE_NAME)
+            .help(ARG_HELP.as_str())
+            .display_order(order)
+    }
+}
+
+/// Handles providing the arg for and retrieval of JSON session and payment args.
```
Fraser999:
```suggestion
/// Handles providing the arg for and retrieval of JSON session args.
```

src/transaction/creation_common.rs
```diff
@@ -0,0 +1,883 @@
...
+
+    fn arg(name: &'static str, order: usize) -> Arg {
+        Arg::new(name)
+            .long(name)
+            .required(false)
+            .value_name(ARG_VALUE_NAME)
+            .help(ARG_HELP)
+            .display_order(order)
+    }
+}
```
Fraser999: I think we should look to drop support for args_complex in this upcoming release.  I'd suggest removing it from the deploy-related code too, but at the very least we shouldn't support it for transactions.  It's really not fit for purpose IMO.

src/transaction/creation_common.rs
```diff
@@ -0,0 +1,883 @@
...
+            .help(ARG_HELP)
+            .display_order(DisplayOrder::PaymentMode as usize)
+    }
+
+    pub fn get(matches: &ArgMatches) -> Option<&str> {
+        matches.get_one::<String>(payment_amount::ARG_NAME).map(String::as_str)
+    }
+}
+
+pub(super) mod initiator_address {
```
Fraser999, line 369: This module should really be merged with the `session_account` one below as `initiator_address` is its replacement.  We eventually want the same functionality in terms if allowing the user to specify a public key file path (hex or pem), a formatted public key, a formatted account hash, a formatted entity address, or nothing (in which case the public key associated with the secret key is used).

lib/lib.rs
```diff
@@ -151,6 +151,19 @@ pub fn output_deploy(output: OutputKind, deploy: &Deploy) -> Result<(), Error> {
     output.commit()
 }
 
+/// Outputs a [`Transaction`] to a file or stdout.
+///
+/// As a file, the `Deploy` can subsequently be signed by other parties using [`sign_transaction_file`]
```
Fraser999:
```suggestion
/// As a file, the `Transaction` can subsequently be signed by other parties using [`sign_transaction_file`]
```

lib/cli/transaction_str_params.rs
Fraser999, line 1: Missed some renames of deploy to transaction.

lib/cli/transaction_str_params.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// A `Deploy` can only be included in a `Block` between `timestamp` and `timestamp + ttl`.
+    /// Input examples: '1hr 12min', '30min 50sec', '1day'.
+    ///
+    /// See [`humantime::parse_duration`] for more information.
+    pub ttl: &'a str,
+    /// Name of the chain, to avoid the `Deploy` from being accidentally or maliciously included in
+    /// a different chain.
+    pub chain_name: &'a str,
+    /// The hex-encoded public key of the account context under which the session code will be
+    /// executed.
```
Fraser999: This is now more than just a public key.  It could be any one of three variants.

src/transaction/creation_common.rs
```diff
@@ -0,0 +1,883 @@
...
+            println!();
+            println!("{}", json_args_help::info_and_examples());
+            return true;
+        }
+
+        false
+    }
+}
+
+pub(super) fn session_str_params(matches: &ArgMatches) -> SessionStrParams<'_> {
```
Fraser999: We could maybe aim to move away from `SessionStrParams` for Transactions, leaving it only for Deploys?

In the same way we currently have `cli::transfer` not taking a general purpose `SessionStrParams`, I think it would be best to have separate functions for each of the `TransactionV1Builder`'s `new_xxx` methods, meaning none of them would need a `SessionStrParams`.

Fraser999 (commented):
A nice start on this epic.

zacshowa:
Some notes for readers of this PR after the recent set of changes:
-

## Parsing of some arguments: 
I chose to put some parsing logic into the modules in creation_common.rs that represent subcommands instead of parse.rs. This might be the wrong location for it and I'm open to changing it, but I was running into some scoping issues and I didn't want to largely change the accessibility without some insight into it.

## Building the transaction builders: 
I build the transaction builder "earlier" than I likely need to, but I think it naturally comes from parsing some arguments from the "nested" subcommands. I could likely change this to package things up as strings initially, to then later parse the args into rust types, and build the transaction builder later in the pipeline with minimal effort.

## End user considerations:
I think it could be beneficial to move all of the arguments to the second layer of subcommands such that the options would all come at the end of the command like this:
 
casper-client make-transaction withdraw-bid --ttl 1day --public-key public key.

Currently some common args are handled by the top level subcommand like this:

casper-client make-transaction --ttl 1day withdraw-bid --public-key public key

## Style nits:
I didn't spend a ton of time going over the new code for style nits. I plan to do that before a final review and the merge of this work.

## Tests: 
I still need to add tests for the transaction creation infrastructure like the ones that already exist for deploys.

lib/cli/parse.rs
```diff
@@ -530,15 +415,6 @@ pub(super) fn payment_executable_deploy_item(
             requires[payment_entry_point] requires_empty[],
         (payment_path) requires[] requires_empty[payment_entry_point, payment_version],
     );
-    if !payment_args_simple.is_empty() && !payment_args_complex.is_empty() {
```
gRoussac, line 533: Maybe you could fetch changes in 1.6 regarding this check ? to see if args simple and json are not both provided ? check_no_conflicting_arg_types

https://github.com/casper-ecosystem/casper-client-rs/blob/dev/lib/cli/parse.rs#L430

lib/cli/transaction_str_params.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// `secret_key`.  It is an error for both fields to be empty.
+    pub initiator_addr: String,
+    /// Simple session args for use in the transaction
+    pub session_args_simple: Vec<&'a str>,
+    /// Session args in json for use with the transaction
+    pub session_args_json: &'a str,
+    /// The pricing mode to use with the transaction
+    pub maybe_pricing_mode: Option<&'a str>,
+    /// The optional output path for the transaction, if writing it to a file.
+    pub maybe_output_path: &'a str,
```
gRoussac: Maybe something suggest that it is an Option, like maybe_pricing_mode, so maybe_output_path being empty is a nice way to tell this empty string will act like None but it's not really a good practice to have maybe_ prefixes for &str.

src/main.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
```
gRoussac: Small question, why is core suddenly required ?

gRoussac (approved):

src/main.rs
zacshowa: I think that was an accidental import that then got missed by clippy.

lib/cli/parse.rs
zacshowa: Given the way the args are constructed via clap, entering a command with both flags will result in an error meaning this both `args_simple` and `args_json` would never both be provided.

zacshowa: I've added a slightly modified version of this function to catch conflicts for the sdk, but to also reflect the removal of args_complex

Issue #121, opened 2023-11-14
Sign Transaction command


RBYGLLS:
Add a sign-transaction command to the client that adds support for signing previously created unsigned transactions similar to the sign-deploy command.

PR #134, opened 2024-02-02
Sign transaction command

zacshowa:
# Summary of work done:
- Add a `sign-transaction` command to enable signing previously created transaction files.

The command specifies the session account twice as I still need to identify and implement the proper inputs via clap arguments
Additionally, the command already supports outputting the transaction to a user specified file location, as long as the directory exists.

```
casper-client  make-transaction transfer --source uref-722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d-001 --target uref-722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d-001 --initiator-address 01722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d --chain-name test -p 10 --transfer-amount 10 --output ~/casper/client-dev/test-transaction
```

Associated output:

```
{
  "hash": "705ed54ff8cfef965cdcd1d1f8f30b4a1bd97d3e25dcf81ab91e786ca2a228e3",
  "header": {
    "chain_name": "test",
    "timestamp": "2024-01-29T20:47:38.961Z",
    "ttl": "30m",
    "body_hash": "fb94fd83178e3acf22546beebf5f44692499d681c4381f6d145d85ff9b5fc152",
    "pricing_mode": {
      "GasPriceMultiplier": 1
    },
    "payment_amount": 10,
    "initiator_addr": {
      "PublicKey": "01722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d"
    }
  },
  "body": {
    "args": [
      [
        "source",
        {
          "cl_type": "URef",
          "bytes": "722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d01",
          "parsed": "uref-722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d-001"
        }
      ],
      [
        "target",
        {
          "cl_type": "URef",
          "bytes": "722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d01",
          "parsed": "uref-722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d-001"
        }
      ],
      [
        "amount",
        {
          "cl_type": "U512",
          "bytes": "010a",
          "parsed": "10"
        }
      ]
    ],
    "target": "Native",
    "entry_point": "Transfer",
    "scheduling": "Standard"
  },
  "approvals": []
}

```

Relevant command to sign the above transaction:
```
casper-client sign-transaction --secret-key ~/casper/keys/client-dev/secret_key.pem --transaction-path ~/casper/client-dev/test-transaction --output ~/casper/client-dev/test-transaction-signed
```

Associated output:
```
{
  "hash": "705ed54ff8cfef965cdcd1d1f8f30b4a1bd97d3e25dcf81ab91e786ca2a228e3",
  "header": {
    "chain_name": "test",
    "timestamp": "2024-01-29T20:47:38.961Z",
    "ttl": "30m",
    "body_hash": "fb94fd83178e3acf22546beebf5f44692499d681c4381f6d145d85ff9b5fc152",
    "pricing_mode": {
      "GasPriceMultiplier": 1
    },
    "payment_amount": 10,
    "initiator_addr": {
      "PublicKey": "01722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d"
    }
  },
  "body": {
    "args": [
      [
        "source",
        {
          "cl_type": "URef",
          "bytes": "722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d01",
          "parsed": "uref-722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d-001"
        }
      ],
      [
        "target",
        {
          "cl_type": "URef",
          "bytes": "722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d01",
          "parsed": "uref-722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d-001"
        }
      ],
      [
        "amount",
        {
          "cl_type": "U512",
          "bytes": "010a",
          "parsed": "10"
        }
      ]
    ],
    "target": "Native",
    "entry_point": "Transfer",
    "scheduling": "Standard"
  },
  "approvals": [
    {
      "signer": "01722e1b3d31bef0ba832121bd2941aae6a246d0d05ac95aa16dd587cc5469871d",
      "signature": "01275c5f4b31eb30d7419820d51d7d23ab5e4860c870401c9f086b2b12aafbc9350ad9aa2be74ef49ce72c079591bf490dce0cc9e2e4caf6c1c4a243ba9e3dbb08"
    }
  ]
}
```

# Associated ticket(s):
- #121 
- [Sign Transaction commnd ](https://app.zenhub.com/workspaces/casperlabs-sdks-64b185bf2cb87924e7759d9d/issues/gh/casper-ecosystem/casper-client-rs/121)

### DIFF
--- a/lib/cli/transaction.rs
+++ b/lib/cli/transaction.rs
@@ -90,6 +90,24 @@ pub fn make_transaction(
     crate::output_transaction(output, &transaction).map_err(CliError::from)
 }
 
+/// Reads a previously-saved [`TransactionV1`] from a file, cryptographically signs it, and outputs it to a
+/// file or stdout.
+///
+/// `maybe_output_path` specifies the output file path, or if empty, will print it to `stdout`.  If
+/// `force` is true, and a file exists at `maybe_output_path`, it will be overwritten.  If `force`
+/// is false and a file exists at `maybe_output_path`, [`Error::FileAlreadyExists`] is returned
+/// and the file will not be written.
+pub fn sign_transaction_file(
+    input_path: &str,
+    secret_key_path: &str,
+    maybe_output_path: &str,
+    force: bool,
+) -> Result<(), CliError> {
+    let output = parse::output_kind(maybe_output_path, force);
+    let secret_key = parse::secret_key_from_file(secret_key_path)?;
+    crate::sign_transaction_file(input_path, &secret_key, output).map_err(CliError::from)
+}
+
 pub fn make_transaction_builder(
     transaction_builder_params: TransactionBuilderParams,
 ) -> Result<TransactionV1Builder, CliError> {
--- a/lib/cli/transaction.rs
+++ b/lib/cli/transaction.rs
@@ -100,10 +100,10 @@ pub fn make_transaction(
 pub fn sign_transaction_file(
     input_path: &str,
     secret_key_path: &str,
-    maybe_output_path: &str,
+    maybe_output_path: Option<&str>,
     force: bool,
 ) -> Result<(), CliError> {
-    let output = parse::output_kind(maybe_output_path, force);
+    let output = parse::output_kind(maybe_output_path.unwrap_or(""), force);
     let secret_key = parse::secret_key_from_file(secret_key_path)?;
     crate::sign_transaction_file(input_path, &secret_key, output).map_err(CliError::from)
 }
--- a/lib/lib.rs
+++ b/lib/lib.rs
@@ -176,6 +176,18 @@ pub fn read_deploy_file<P: AsRef<Path>>(deploy_path: P) -> Result<Deploy, Error>
     read_deploy(Cursor::new(input))
 }
 
+/// Reads a previously-saved [`Transaction`] from a file.
+pub fn read_transaction_file<P: AsRef<Path>>(transaction_path: P) -> Result<TransactionV1, Error> {
+    let input = fs::read(transaction_path.as_ref()).map_err(|error| Error::IoError {
+        context: format!(
+            "unable to read transaction file at '{}'",
+            transaction_path.as_ref().display()
+        ),
+        error,
+    })?;
+    read_transaction(Cursor::new(input))
+}
+
 /// Reads a previously-saved [`Deploy`] from a file, cryptographically signs it, and outputs it
 /// to a file or stdout.
 ///
@@ -195,6 +207,24 @@ pub fn sign_deploy_file<P: AsRef<Path>>(
     deploy.is_valid_size(MAX_SERIALIZED_SIZE_OF_DEPLOY)?;
 
     write_deploy(&deploy, output.get()?)?;
+    output.commit()
+}
+
+/// Reads a previously-saved [`TransactionV1`] from a file, cryptographically signs it, and outputs it to a file or stdout.
+///
+/// `output` specifies the output file and corresponding overwrite behaviour, or if OutputKind::Stdout,
+/// causes the `Transaction` to be printed `stdout`.
+///
+pub fn sign_transaction_file<P: AsRef<Path>>(
+    input_path: P,
+    secret_key: &SecretKey,
+    output: OutputKind,
+) -> Result<(), Error> {
+    let mut transaction = read_transaction_file(input_path)?;
+
+    transaction.sign(secret_key);
+
+    write_transaction(&transaction, output.get()?)?;
     output.commit()
 }
 
@@ -520,9 +550,9 @@ fn write_deploy<W: Write>(deploy: &Deploy, mut output: W) -> Result<(), Error> {
         })
 }
 
-fn write_transaction<W: Write>(deploy: &TransactionV1, mut output: W) -> Result<(), Error> {
+fn write_transaction<W: Write>(transaction: &TransactionV1, mut output: W) -> Result<(), Error> {
     let content =
-        serde_json::to_string_pretty(deploy).map_err(|error| Error::FailedToEncodeToJson {
+        serde_json::to_string_pretty(transaction).map_err(|error| Error::FailedToEncodeToJson {
             context: "writing transaction",
             error,
         })?;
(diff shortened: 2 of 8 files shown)


lib/lib.rs
```diff
@@ -176,6 +176,18 @@ pub fn read_deploy_file<P: AsRef<Path>>(deploy_path: P) -> Result<Deploy, Error>
     read_deploy(Cursor::new(input))
 }
 
+/// Reads a previously-saved [`Transaction`] from a file.
+pub fn read_transaction_file<P: AsRef<Path>>(transaction_path: P) -> Result<TransactionV1, Error> {
```
zajko, line 180: does this need to be public?

src/transaction/sign.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    async fn run(matches: &ArgMatches) -> Result<Success, CliError> {
+        let input_path = creation_common::transaction_path::get(matches);
+        let secret_key = common::secret_key::get(matches).unwrap_or_default();
+        let maybe_output_path = creation_common::output::get(matches).unwrap_or_default();
+        let force = common::force::get(matches);
+        casper_client::cli::sign_transaction_file(input_path, secret_key, maybe_output_path, force)
+            .map(|_| {
+                Success::Output(if maybe_output_path.is_empty() {
```
zajko: Since this is already in a block expression, could we move the string calculation before the Success::Output? Like
```
let output = if maybe_output_path.is_empty() {
String::new()
                } else {
                    format!(
                        "Signed the transaction at {} and wrote to {}",
                        input_path, maybe_output_path
                    )
                }
Success::Output(output)
```

zajko (approved):

lib/lib.rs
zacshowa: I gave this function the same level of exposure as the other functions here, similar to the read_deploy_file function. I would argue it needs to be public as the lib should support this functionality. Additionally, the client itself needs to pull this function from the lib.

src/transaction/sign.rs
zacshowa: I think that this could make the function more readable in general. 
An alternative form that I ended up with looked like this to keep the functions return signature the same:

```
        let output = if maybe_output_path.is_empty(){
            String::new()
        } else {
            format!(
                "Signed the transaction at {} and wrote to {}",
                input_path, maybe_output_path
            )
        };
        casper_client::cli::sign_transaction_file(input_path, secret_key, maybe_output_path, force)
            .map(|_| {
                Success::Output(output)
            })
```

This way we keep the `Result<Success, CliError>`  structure on the last line via map.
I'm okay with either approach.  @gRoussac care to weigh in?

lib/lib.rs
zajko: :+1:

src/transaction/sign.rs
gRoussac: Hi, well I am not disturbed by the first version, except indentation, just a small new line make it a bit clearer imo. Otherwise I am am also of course fine to prepare an output var.

```
.map(|_| {
    Success::Output(
        if maybe_output_path.is_empty() {
            String::new()
        } else {
            format!(
                "Signed the transaction at {} and wrote to {}",
                input_path, maybe_output_path
            )
        }
    )
})
```
```
let output = if maybe_output_path.is_empty() {
    String::new()
} else {
    format!(
        "Signed the transaction at {} and wrote to {}",
        input_path, maybe_output_path
    )
};

casper_client::cli::sign_transaction_file(input_path, secret_key, maybe_output_path, force)
    .map(|_| Success::Output(output))
    
```

gRoussac (approved):

lib/cli/transaction.rs
```diff
@@ -90,6 +90,24 @@ pub fn make_transaction(
...
+/// file or stdout.
+///
+/// `maybe_output_path` specifies the output file path, or if empty, will print it to `stdout`.  If
+/// `force` is true, and a file exists at `maybe_output_path`, it will be overwritten.  If `force`
+/// is false and a file exists at `maybe_output_path`, [`Error::FileAlreadyExists`] is returned
+/// and the file will not be written.
+pub fn sign_transaction_file(
+    input_path: &str,
+    secret_key_path: &str,
+    maybe_output_path: &str,
```
gRoussac: You don't need to specify maybe_xxx or make it an Option in this case, relying on an empty string does not read very well on consumer point of view 

casper_client::cli::sign_transaction_file(input_path, secret_key, "", true). 

It's also a bit strange to have a mandatory input path and a non mandatory output path but I guess this will be okay for now.

zacshowa: I agree that relying on a empty string isn't ideal. The parameter should be optional as the default behavior of the client when no output path is specified is to output to stdout. I would argue that, as optional parameter, it should retain the name `maybe_output_path`. However, when I went to change this to an `Option<&str>` along with changing the parsing behavior, I realized that functions related to deploys also depend on the same parsing function. Therefore, I would need to change the API of previously exposed library functions, or write specific parsing code for `OutputKind` that is unique to transactions. The client should, In my opinion, maintain consistent behavior between the two. Maybe this switch to options would be better served in a different PR?

gRoussac: Of course I can live with current code but I would start wrapping that maybe pattern where ever I can in the API, if I understand correctly user is not using  crate::sign_transaction_file directly but this interface so I would mask this little empty string trick

I suggest thus, but not a perquisite 

```
pub fn sign_transaction_file(
    input_path: &str,
    secret_key_path: &str,
    maybe_output_path: Option<&str>,
    force: bool,
) -> Result<(), CliError> {
    let output = parse::output_kind(maybe_output_path.unwrap_or(""), force);
    let secret_key = parse::secret_key_from_file(secret_key_path)?;
    crate::sign_transaction_file(input_path, &secret_key, output).map_err(CliError::from)
}
```

Cheers

zacshowa: I ended up going with this suggestion as it's a good interim solution. I'll see if it would be worth creating a ticket to make this uniform across the client.

PR #135, opened 2024-02-02
Add support for addressable entity requests

jacek-casper:
We will need to support a new endpoint that allows retrieval of addressable entities with fallback for legacy accounts.
The endpoint has been added to the RPC sidecar in this PR: https://github.com/zajko/event-sidecar/pull/10.
There's a test for it in casper-test which uses the code added here: https://github.com/casper-network/casper-test/pull/146


zacshowa (approved):
The only addition I would like to see is adding this change to CHANGELOG.MD. Otherwise, LGTM!

PR #136, opened 2024-02-13
Patch to address updated signature of `Key::AddressableEntity`

zacshowa:
Recent changes to `casper-types::Key::AddressableEntity` were causing the casper-client to fail to compile for nctl-test
This commit updates the client to support the new `AddressableEntity` signature


AlexanderLimonov:
This looks good. This is a potential blocker for a bunch of PRs about to drop, so I'm merging it now.

Issue #127, opened 2023-11-14
Add clap command for put-transaction RPC method


RBYGLLS:
This ticket will be for adding another command to the client via clap, it will utilize the the work done to create transactions and include creating new code to craft and send the RPC request to a node.

PR #137, opened 2024-02-13
Put transaction command

zacshowa:
# Summary of work done:
- Add a `put-transaction` command to enable sending transactions to casper networks via JSON-RPC

# Areas of note for reviewers:
- the location of arguments in the command. As the top level command needs to have several args that it previously did not, there might be some desire to migrate the `node-address` and `rpc-id` args to the inner subcommands and optionally pass it back to the main command.
- I had to silence some clippy warnings in v1_6_0.rs and v1_5_0.rs as some warnings arose related to unused result exports. I'm under the impression these are still worth exporting until the client can handle multiple different node versions. 

```
casper-client put-txn --node-address http://localhost:11101/rpc session --transaction-path ~/casper/casper-checkpoint/checkpoint_contract/contract/target/wasm32-unknown-unknown/release/contract.wasm --session-entry-point test --chain-name casper-net-1 --secret-key ~/casper/casper-node/utils/nctl/assets/net-1/users/user-1/secret_key.pem --payment-amount 10

```

Associated output:

```
{
  "jsonrpc": "2.0",
  "id": 2007083436180091612,
  "result": {
    "api_version": "1.0.0",
    "transaction_hash": {
      "Version1": "16148cbb254970a85e6df685d721606709167a602915e7604d7f92682f2683fb"
    }
  }
}

```

# Associated ticket(s):
- #127 
- [Put Transaction command ](https://app.zenhub.com/workspaces/casperlabs-sdks-64b185bf2cb87924e7759d9d/issues/gh/casper-ecosystem/casper-client-rs/127)

### DIFF
--- a/lib/cli/transaction.rs
+++ b/lib/cli/transaction.rs
@@ -1,16 +1,20 @@
 use crate::cli::{parse, CliError, TransactionBuilderParams, TransactionStrParams};
-use casper_types::{InitiatorAddr, TransactionSessionKind, TransactionV1, TransactionV1Builder};
+use crate::rpcs::results::PutTransactionResult;
+use crate::SuccessResponse;
+use casper_types::{
+    InitiatorAddr, Transaction, TransactionSessionKind, TransactionV1, TransactionV1Builder,
+};
 
 pub fn create_transaction(
     builder_params: TransactionBuilderParams,
     transaction_params: TransactionStrParams,
-    allow_unsigned_deploy: bool,
+    allow_unsigned_transaction: bool,
 ) -> Result<TransactionV1, CliError> {
     let chain_name = transaction_params.chain_name.to_string();
     if transaction_params.payment_amount.is_empty() {
         return Err(CliError::InvalidArgument {
             context: "create_transaction (payment_amount)",
-            error: "payment_amount is required".to_string(),
+            error: "payment_amount is required to be non empty".to_string(),
         });
     }
     let payment_amount = transaction_params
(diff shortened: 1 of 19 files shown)


gRoussac (approved):

lib/cli/transaction.rs
```diff
@@ -90,6 +94,31 @@ pub fn make_transaction(
...
+    rpc_id_str: &str,
+    node_address: &str,
+    verbosity_level: u64,
+    builder_params: TransactionBuilderParams<'_>,
+    transaction_params: TransactionStrParams<'_>,
+) -> Result<SuccessResponse<PutTransactionResult>, CliError> {
+    let rpc_id = parse::rpc_id(rpc_id_str);
+    let verbosity_level = parse::verbosity(verbosity_level);
+    let transaction = create_transaction(builder_params, transaction_params, false)?;
+    crate::put_transaction(
```
zajko: I'm not a huge fan of path qualified calls. Out of curiosity - why not use an alias for `put_transaction` and import it on top of the file?

src/transaction/creation_common.rs
```diff
@@ -522,21 +524,18 @@ pub(super) mod transaction_path {
     const ARG_VALUE_NAME: &str = common::ARG_PATH;
     const ARG_HELP: &str = "Path to input transaction file";
 
-    pub fn arg() -> Arg {
+    pub fn arg(required: bool) -> Arg {
```
zajko: I see this called as `arg(true)`, but I don't see `arg(false)`. Is there a point in having this as a parameter if it's always true?

zajko (approved):

PR #138, opened 2024-02-20
Various fixes to address updates to casper-types

zacshowa:
As changes have been merged into the feat-2.0 branch of casper-types, the client has had to update some type signatures to keep up to date. This set of changes is the latest installment in that saga.


AlexanderLimonov (approved):
Looks good, let's try it

PR #142, opened 2024-03-26
Update dictionary item request with entity variant

jacek-casper:
Adds a new version of the dictionary item request with a new variant for entities to match parity with what we already support for legacy accounts/contracts.

Depends on:
- https://github.com/casper-network/casper-node/pull/4628
- https://github.com/casper-network/casper-sidecar/pull/267


jacek-casper:
@rafal-ch @zacshowa all the upstream stuff has been merged, so this is ready to review and merge

zacshowa (approved):
LGTM

PR #143, opened 2024-03-29
Changes required to work with the latest version of node 2.0

rafal-ch:


lib/cli/parse.rs
```diff
@@ -782,30 +781,37 @@ pub(super) fn entity_identifier(entity_identifier: &str) -> Result<EntityIdentif
...
-    maybe_paid_amount: Option<U512>,
+    maybe_paid_amount: Option<u64>,
 ) -> Result<PricingMode, CliError> {
     match pricing_mode_identifier_str.to_lowercase().as_str() {
         "classic" => {
-            if maybe_gas_price_str.is_empty() {
+            if maybe_gas_price_tolerance_str.is_empty() {
                 return Err(CliError::InvalidArgument {
                     context: "gas-price",
                     error: "Gas price is required".to_string(),
```
alsrdn: nit: maybe we should change the message here too: `Gas price tolerance required`

src/transaction/creation_common.rs
```diff
@@ -355,8 +358,57 @@ pub(super) mod paid_amount {
...
+    }
+}
+
+pub(super) mod standard_payment {
+    use super::*;
+    pub(in crate::transaction) const ARG_NAME: &str = "standard-payment";
+
+    const ARG_VALUE_NAME: &str = common::ARG_STRING;
+
+    const ARG_HELP: &str = "Standard payment.";
```
alsrdn: nit: a more helpful help string would be nice here. Something like: `Flag to determine if this transaction uses standard or custom payment`.

alsrdn (approved):

jacek-casper (approved):

lib/cli/parse.rs
rafal-ch: Updated in 44a0e28c611d6623c8a6ac2ccafa3f4d27cc5131

src/transaction/creation_common.rs
rafal-ch: Updated in 041470958d8151e2cb405baffe724234ddc591b9

PR #144, opened 2024-04-05
Interaction with Casper smart-contract source code verification service

### LINKED ISSUE #187: Add Source Code Verification option
Description: Add a capability to interact with the Casper Smart Contract Source Code Verification Service(SCVS). Include commands and functions for verifying contracts, parsing arguments, creating archives, handling errors, logging, and interfacing with the verification API. Please refer to SCVS repo [here](https://github.com/teonite/Casper-SCVS-Verificator), for more details.

dev/2.0

moubctez:
Allow interaction with Source Code Verification Service.

Check https://github.com/teonite/Casper-SCVS-Verificator for details.


lib/verification.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    let Ok(http_client) = ClientBuilder::new()
+        .default_headers(headers)
+        // https://github.com/hyperium/hyper/issues/2136
+        .pool_max_idle_per_host(0)
+        .user_agent("casper-client-rs")
+        .build()
+    else {
+        eprintln!("Failed to build HTTP client");
+        return Err(Error::ContractVerificationFailed); // FIXME: different error
```
zacshowa: I think something like `Error::FailedToConstructHttpClient` would suffice here.

lib/verification.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    match response.status() {
+        StatusCode::OK => {
+            if verbosity == Verbosity::Medium || verbosity == Verbosity::High {
+                println!("Sent verification request",);
+            }
+        }
+        status => {
+            eprintln!("Verification failed with status {status}");
+            // return Err(Error::ContractVerificationFailed);
```
zacshowa: Any extraneous comments should be removed.

lib/verification_types.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub(crate) result: VerificationDetailsResult,
+}
+
+// #[derive(Serialize)]
+// pub(crate) struct Toolchain {
+//     pub(crate) channel: String,
+//     pub(crate) components: Vec<String>,
+//     pub(crate) targets: Vec<String>,
+//     pub(crate) profile: String,
+// }
```
zacshowa: This commented struct, as well as any references to it in other structs, e.g. `VerificationDetailsResult`, should be removed until the struct is used.

lib/verification.rs
```diff
@@ -0,0 +1,208 @@
...
+                }
+            }
+            Err(error) => {
+                eprintln!("Cannot get verification status: {error:?}");
+                break;
+            }
+        };
+
+        retries -= 1;
+    }
```
zacshowa: As this function is indirectly exposed via lib.rs, we might do well to have the number of retries/delay configurable by the caller of send_verification_request.

Otherwise we should, at the very least, make `retries` and `delay` constants at the top of the file, rather than pseudo-magic numbers defined in the function.

lib/verification.rs
```diff
@@ -0,0 +1,208 @@
...
+/// # Returns
+///
+/// The compressed tar archive as a `Bytes` object, or an `std::io::Error` if an error occurs during
+/// the archiving process.
+pub fn build_archive(path: &Path) -> Result<Bytes, std::io::Error> {
+    let buffer = BytesMut::new().writer();
+    let encoder = GzEncoder::new(buffer, Compression::best());
+    let mut archive = TarBuilder::new(encoder);
+
+    for entry in (path.read_dir()?).flatten() {
```
zacshowa: Just a style nit
```suggestion
    for entry in path.read_dir()?.flatten() {
```

lib/lib.rs
```diff
@@ -554,3 +561,44 @@ pub async fn get_era_info(
...
+    key: Key,
+    verification_url_base_path: &str,
+    verbosity: Verbosity,
+) -> Result<VerificationDetails, Error> {
+    if verbosity == Verbosity::Medium || verbosity == Verbosity::High {
+        println!("Key: {key}");
+        println!("Verification service base path: {verification_url_base_path}",);
+    }
+
+    let project_path = match current_dir() {
```
zacshowa: Wouldn't it be better to allow the user to pass in some `target_dir` for the client to scan?

zacshowa (changes requested):
Largely looks good.

I would also like to see some discussion surrounding the retry logic in this code.

There are some style nits to address, and this change should be added to the change log entry corresponding to the release that you want this code to go out in.

src/verify_contract.rs
```diff
@@ -0,0 +1,73 @@
...
+            .get_one::<String>(ARG_NAME)
+            .map(String::as_str)
+            .unwrap_or_else(|| panic!("should have {ARG_NAME} arg"))
+    }
+}
+
+#[async_trait]
+impl ClientCommand for VerifyContract {
+    const NAME: &'static str = "verify-contract";
+    const ABOUT: &'static str = "Verify smart contract source code";
```
zacshowa: One additional comment, I would like to see a more robust help message that describes the behavior of this command for end users.

src/verify_contract.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    pub fn get(matches: &ArgMatches) -> Option<&str> {
+        matches.get_one::<String>(ARG_NAME).map(String::as_str)
+    }
+}
+
+#[async_trait]
+impl ClientCommand for VerifyContract {
+    const NAME: &'static str = "verify-contract";
+    const ABOUT: &'static str = "Verify smart contract source code using verification service; the source code will be upload, built and compared against the deployed contract binary";
```
zacshowa:
```suggestion
    const ABOUT: &'static str = "Verifies a smart contracts source code using verification service. 
    The source code will be uploaded, built, and compared against the deployed contract binary.
    You may specify a path from which the code will be read and compressed from, or omit the path.
    If the path is omitted, the archive will be built from the current working directory.";
```

zacshowa (approved):
I have one additional suggestion for the about message for the command, Other than that this looks good to me.

moubctez:
> I have one additional suggestion for the about message for the command, Other than that this looks good to me.

Good suggestion. I've applied it.

zacshowa:
Just wanted to keep you up to date on this PR. Currently it's failing Cargo audit because of a dependency having a security issue,
This is an issue that is currently being fixed in [this PR](https://github.com/casper-ecosystem/casper-client-rs/pull/149).

Once that PR merges, you should be able to cherry pick the changes into this PR and make it ready to merge.

moubctez:
> Just wanted to keep you up to date on this PR. Currently it's failing Cargo audit because of a dependency having a security issue, This is an issue that is currently being fixed in [this PR](https://github.com/casper-ecosystem/casper-client-rs/pull/149).
> 
> Once that PR merges, you should be able to cherry pick the changes into this PR and make it ready to merge.

I have upgraded "vergen" to v8 to resolve the issue.

lib/verification.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+/// Verifies the smart contract code against the one deployed at deploy hash.
+///
+/// Sends a verification request to the specified verification URL base path, including the deploy hash,
+/// public key, and code archive.
+///
+/// # Arguments
+///
+/// * `deploy_hash` - The hash of the deployed contract.
```
gRoussac: Hi, The code documentation does not seem to follow function params below.

gRoussac (changes requested):

Cargo.toml
```diff
@@ -28,10 +28,13 @@ std-fs-io = ["casper-types/std-fs-io"]
 [dependencies]
 async-trait = { version = "0.1.74", optional = true }
 base16 = "0.2.1"
+base64 = "0.21.5"
+bytes = "1.5.0"
```
gRoussac: Would it make sense to use capser-types::Bytes instead adding a library ?

Cargo.toml
```diff
@@ -28,10 +28,13 @@ std-fs-io = ["casper-types/std-fs-io"]
 [dependencies]
 async-trait = { version = "0.1.74", optional = true }
 base16 = "0.2.1"
+base64 = "0.21.5"
```
gRoussac: Could we please restrict this import to the minimum required ?

base64 = { version = "0.22.1", default-features = false }

or maybe

base64 = { version = "0.22.1", default-features = false, features = ["alloc"] }

Cargo.toml
```diff
@@ -28,10 +28,13 @@ std-fs-io = ["casper-types/std-fs-io"]
...
+bytes = "1.5.0"
 casper-hashing = "3.0.0"
 casper-types = { version = "4.0.1", features = ["std"] }
 clap = { version = "~4.4", optional = true, features = ["cargo", "deprecated", "wrap_help"] }
 clap_complete = { version = "<4.5.0", optional = true }
+flate2 = "1.0.28"
```
gRoussac: Is it possible to remove the default feature and still work ? thanks

Cargo.toml
```diff
@@ -43,6 +46,7 @@ reqwest = { version = "0.12.3", features = ["json"] }
 schemars = "=0.8.5"
 serde = { version = "1.0.193", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.108", features = ["preserve_order"] }
+tar = "0.4.40"
```
gRoussac: Is it possible to remove the default feature and still work ? thanks

lib/verification.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+/// Waits for the verification process to finish.
+async fn wait_for_verification_finished(
+    base_url: &str,
+    http_client: &Client,
+    key: Key,
+    verbosity: Verbosity,
+) {
+    let mut retries = MAX_RETRIES;
```
gRoussac, line 165: Are there tests that cover this logic of retrying 1200 times every 3 seconds ? is there a way to limit this load ?

10 attempts seems more reasonable with time extending between attempts

const MAX_RETRIES: u32 = 10;
const BASE_DELAY: u64 = 3;
const MAX_DELAY: u64 = 300;

```delay = std::cmp::min(delay * 2, MAX_DELAY); ```

Would this be possible to unit test this function ?

File: Cargo.toml (hunk 10 above)
Comment (moubctez) [reply Cargo.toml]: No, it seems to be a different type.

File: Cargo.toml (hunk 12 above)
Comment (moubctez) [reply Cargo.toml]: Unfortunately, not.

File: Cargo.toml (hunk 13 above)
Comment (moubctez) [reply Cargo.toml]: Yes, it is. We don't need xattrs.

File: build.rs [hunk 15]
```diff
@@ -1,21 +0,0 @@
-use std::process::Command;
```
gRoussac: Hi @moubctez, cc @devendran-m thanks for treating my review.

Sorry I am missing some info here, was there a decision to remove that build code ? because I just opened a [PR](https://github.com/casper-ecosystem/casper-client-rs/pull/175) to condition that git versioning.

this `GIT_SHA_SHORT` is used  in main.rs and I don't see any change regarding this in this PR, I am missing something ? thanks

gRoussac (changes requested):

build.rs
moubctez: It must have been killed in action of conflict resolution. I've restored `build.rs`.

gRoussac (approved):

build.rs
gRoussac: Hi, Sorry but why are you reverting to vergen ?

This file has evolved I believe vergen was removed 

https://github.com/casper-ecosystem/casper-client-rs/blob/dev/build.rs

from https://github.com/casper-ecosystem/casper-client-rs/pull/149

gRoussac (changes requested):

moubctez:
@gRoussac Now `build.rs` is the correct one. I had ancient history in `dev`. My fault.

gRoussac (approved):

gRoussac:
@moubctez You build is failing tokio is an optional dep so it needs to be specified when used
try to add that arround 
```
#[cfg(feature = "tokio")]
```
https://github.com/casper-ecosystem/casper-client-rs/pull/144/commits/44fef2de9f05ae08ca0c7562410fd68ce2a6859c#file-lib-verification-rs-L11

probably an import issue for second error 

https://github.com/casper-ecosystem/casper-client-rs/pull/144/commits/44fef2de9f05ae08ca0c7562410fd68ce2a6859c#file-lib-lib-rs-L577

Thanks

gRoussac:
When/if merging please choose, "**Squash and merge**" thank you

devendran-m:
Hi @gRoussac , could we now merge this PR please?

PR #145, opened 2024-04-09
Use EntityAddr instead of AddressableEntityHash

jacek-casper:
The initial code was written before `EntityAddr` was introduced, so we didn't use it, but now that `EntityAddr` exists, it should be used here instead for consistency with other code (like for example `GetDictionaryItem`).

Depends on https://github.com/casper-network/casper-sidecar/pull/275 and https://github.com/casper-network/casper-node/pull/4653.


zacshowa (approved):
LGTM

PR #146, opened 2024-04-10
Fix casper types dependency

jacek-casper:
The version has been bumped in the node, which leads to build failures, because now casper client ends up trying to get it from crates.io. This aligns the version with current casper-node: https://github.com/casper-network/casper-node/blob/feat-2.0/types/Cargo.toml#L3.


fizyk20 (approved):

zacshowa (approved):

mgr-inz-rafal (approved):
LGTM :+1:

PR #149, opened 2024-04-15
Address `dev` branch toolchain issues.

zacshowa:
# Changes in this PR

1. This PR adds a `rust-toolchain.rs` to the repo, pinning the rust toolchain to the same `dev` in casper-node.

   - To achieve this, the PR pins 2 clap dependencies to versions that compile on the previously mentioned rust-toolchain.

2. Additionally, this PR removes a dependency on `vergen` in the `build.rs` build script.
   - This functionality is maintained by replacing `vergen` with standard library functions to ask Cargo to populate an env var during the build lifecycle that the client reads while constructing its version string.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,6 @@ readme = "README.md"
 homepage = "https://casperlabs.io"
 repository = "https://github.com/casper-ecosystem/casper-client-rs"
 license = "Apache-2.0"
-build = "build.rs"
 
 [lib]
 crate-type = ["cdylib", "rlib"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,8 @@ async-trait = { version = "0.1.74", optional = true }
 base16 = "0.2.1"
 casper-hashing = "3.0.0"
 casper-types = { version = "4.0.1", features = ["std"] }
-clap = { version = "=4.4.10", optional = true, features = ["cargo", "deprecated", "wrap_help"] }
-clap_complete = { version = "=4.4.4", optional = true }
+clap = { version = "~4.4", optional = true, features = ["cargo", "deprecated", "wrap_help"] }
+clap_complete = { version = "<4.5.0", optional = true }
 hex-buffer-serde = "0.4.0"
 humantime = "2.1.0"
 itertools = "0.12.0"
@@ -49,8 +49,6 @@ uint = "0.9.5"
 
 [dev-dependencies]
 tempfile = "3.8.1"
-
-[build-dependencies]
 
 [patch.crates-io]
 casper-hashing = { git = "https://github.com/casper-network/casper-node", branch = "dev" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -51,7 +51,6 @@ uint = "0.9.5"
 tempfile = "3.8.1"
 
 [build-dependencies]
```
gRoussac: I guess you can remove [build-dependencies] now.

Are you really really sure you have to fix clap ?

marc-casperlabs: At least I observed issues where `clap_lex` (and likely other clap packages) specified a minimum Rust version (the [is already on 1.74](https://github.com/clap-rs/clap/blob/0cd10d19f0ac39341f89130d924e87879b58867f/Cargo.toml#L17)).

You can see that [clap 4.4.18 fixes it at 1.70](https://github.com/clap-rs/clap/blob/0134f45ff0e2e2be8c451565e4fbf5d3cb7b7cfd/Cargo.toml#L17), whereas [4.5.0 already requires 1.74](https://github.com/clap-rs/clap/blob/a751c5fe65cd33cb09e85ff3039b4fd0182cdb6e/Cargo.toml#L17).

We could consider something less strict like `<4.5.0` instead of `=...` though?

marc-casperlabs:
Looks good, although consider the slightly more liberal version fix. @zacshowa I think getting rid of `vergen` might also be something to do on the main `casper-node`, what do you think?

cc @rafal-ch @alsrdn @jacek-casper

Cargo.toml
zacshowa: I agree that the more liberal we can be with the versioning the better. I don't know what the SOP is for moving the toolchain version forward on the node. That is, when should we be moving the versioning forward vs leaving it the same.  

Seeing as we currently don't use any of these cutting edge `clap` features, I don't see any reason to increment the toolchain version as we don't have a need to do so on the client.

zacshowa:
> I think getting rid of `vergen` might also be something to do on the main `casper-node`

I agree, It's functionality could easily be replaced with standard library functions and calls to git. This would reduce the future risk of transitive dependency security issues at the cost of a small effort.

Cargo.toml
zacshowa: As it turns out the best I can do on this is pinning `clap` to `~4.4` and putting `clap_complete` at `<4.5.0`.

Setting `clap` at `<4.5`, for some esoteric reason, results in issues with `clap_lex` specifying Rust version 1.74

marc-casperlabs (approved):
Looks good!

ryoknz:
FYI, [`actions-rs/*`](https://github.com/actions-rs) actions are deprecated, please consider using https://github.com/dtolnay/rust-toolchain or follow https://github.com/rust-lang/rust-analyzer/blob/99718d0c8bc5aadd993acdcabc1778fc7b5cc572/.github/workflows/ci.yaml#L63-L67
Check https://github.com/rust-lang/rustup/issues/3409

PR #152, opened 2024-05-02
Remove skip_serializing_if attribute to align with the RPC

jacek-casper:
The attribute was deleted in the RPC recently, and now the client unexpectedly gets a null for a field it expects to be omitted


zacshowa (approved):
LGTM

PR #153, opened 2024-05-06
Complete feature std-fs-io in casper-client feat-2.0

### LINKED ISSUE #84: std-fs-io in casper-client feat-2.0 / dev
no-std is not required for web target

gRoussac:
Add wasm32-unknown-unknown target

Add default fetaure

Add `std-fs-io`

````
cargo build --lib --target wasm32-unknown-unknown --no-default-features
````

Remove filesystem I/O functionality from the `std` feature, and gate this behind a new feature `std-fs-io` which depends upon std.

Allows to compile types without `secret_key_from_file/read_transaction_file/make_transaction/send_transaction_file/speculative_send_transaction_file/sign_transaction_file/json_pretty_print/....`

Update crates 

Add `with_bytes()` / `session_bytes` to `SessionStrParams `payment_bytes` to `PaymentStrParams`

Pub `struct JsonArg`

Mutualize code for `get_maybe_secret_key()`

Add `lint-no-default-features/check-no-default-features/keygen::run`

Add documentation on `args_from_simple_or_json`

Add some Clone on exposed result types

Change tokio to `[tokio::main(flavor = "current_thread")]` as dev branch (required)

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -11,49 +15,69 @@ repository = "https://github.com/casper-ecosystem/casper-client-rs"
...
 [dev-dependencies]
-tempfile = "3.7.1"
+tempfile = "3.8.1"
 
 [build-dependencies]
 vergen = { version = "7", default-features = false, features = ["git"] }
 
 [patch.crates-io]
-casper-types = { git = "https://github.com/casper-network/casper-node", branch = "feat-2.0" }
+casper-types = { git = "https://github.com/casper-network/casper-node", branch = "rustSDK-feat-2.0" }
```
gRoussac: This needs to be removed before merge, after casper-typs [PR](https://github.com/casper-network/casper-node/pull/4714)

src/main.rs
```diff
@@ -166,7 +166,7 @@ fn cli() -> Command {
         ))
 }
 
-#[tokio::main]
+#[tokio::main(flavor = "current_thread")]
```
zacshowa: Was there a particular reason to change this? I feel like I might have asked this previously if I've seen this change, but I don't understand why we would want to try to coerce tokio into executing this code on only a single thread. It's usually pretty intelligent about deciding how to handle parallelization under the hood.

zacshowa (approved):
Overall this LGTM.
I just had a small discussion point around the `tokio` runtime flavor change. Mainly for my own curiosity.

src/main.rs
gRoussac: Fraser found out that the client was not multi thread and removed "rt-multi-thread feature flag" on dev

quoting him 
```
I have no idea why I decided on the multi-threaded runtime for tokio.  There's no need for it in the client
```

https://github.com/casper-ecosystem/casper-client-rs/pull/153/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L52

``` tokio can efficiently handle many concurrent I/O operations even on a single thread due to its asynchronous nature.```

I believe until the client does not make simultaneous requests or many while waiting for RPC responses the client does not need this dependency.

File: Cargo.toml [hunk 3]
```diff
@@ -11,58 +11,64 @@ repository = "https://github.com/casper-ecosystem/casper-client-rs"
 license = "Apache-2.0"
 
 [lib]
+crate-type = ["cdylib", "rlib"]
```
jacek-casper, line 18: is the cdylib needed for wasm?

gRoussac: yes indeed

lib/rpcs/v2_0_0/get_entity.rs
```diff
@@ -21,22 +22,43 @@ pub enum EntityIdentifier {
...
-        entry_points: Vec<EntryPointValue>,
-    },
+    AddressableEntity(AddressableEntity),
     /// A legacy account.
     LegacyAccount(Account),
 }
 
+impl EntityOrAccount {
+    /// Returns the addressable entity if present.
+    pub fn addressable_entity(&self) -> Option<AddressableEntity> {
```
jacek-casper: I'd probably change this to return a reference and let the user decide if they want to clone or not
```suggestion
    pub fn addressable_entity(&self) -> Option<&AddressableEntity> {
```

lib/rpcs/v2_0_0/get_entity.rs
```diff
@@ -21,22 +22,43 @@ pub enum EntityIdentifier {
...
+        if let EntityOrAccount::AddressableEntity(addressable_entity) = &self {
+            Some(addressable_entity.clone())
+        } else {
+            None
+        }
+    }
+
+    /// Returns the legacy account if present.
+    pub fn legacy_account(&self) -> Option<Account> {
```
jacek-casper: same as above

lib/cli/transaction.rs
```diff
@@ -94,22 +92,28 @@ pub fn create_transaction(
...
-    let output = parse::output_kind(transaction_params.output_path, force);
-    let transaction = create_transaction(builder_params, transaction_params, true)?;
-    crate::output_transaction(output, &transaction).map_err(CliError::from)
+    #[allow(unused_variables)] force: bool,
+) -> Result<TransactionV1, CliError> {
+    let transaction = create_transaction(builder_params, transaction_params.clone(), true)?;
+    #[cfg(feature = "std-fs-io")]
+    {
+        let output = parse::output_kind(transaction_params.output_path, force);
+        let _ = crate::output_transaction(output, &transaction).map_err(CliError::from);
```
jacek-casper:
```suggestion
        crate::output_transaction(output, &transaction).map_err(CliError::from)?;
```
we shouldn't ignore the error here

lib/cli/tests.rs
```diff
@@ -399,11 +399,13 @@ fn should_fail_to_create_transfer_with_no_secret_key_while_not_allowing_unsigned
...
+    let _error_string =
+        "allow_unsigned_deploy was false, but no secret key was provided".to_string();
     assert!(matches!(
         transfer_deploy.unwrap_err(),
         CliError::InvalidArgument {
-            context: "new_transfer (secret_key, allow_unsigned_deploy)",
-            error: _
+            context: "new_transfer",
+            error: _error_string,
         }
```
jacek-casper: this is not matching on the actual value of the `_error_string`, it's just binding a new `_error_string` variable, this should either match on the actual string literal or use an `if` guard to the match:
`matches!(..., CliError::InvalidArgument {..., error} if error == expected`

lib/cli/tests.rs
```diff
@@ -421,11 +423,13 @@ fn should_fail_to_create_deploy_with_payment_and_session_with_no_secret_key_whil
...
     assert!(transfer_deploy.is_err());
+    let _error_stirng =
+        "allow_unsigned_deploy was false, but no secret key was provided".to_string();
     assert!(matches!(
         transfer_deploy.unwrap_err(),
         CliError::InvalidArgument {
-            context: "with_payment_and_session (secret_key, allow_unsigned_deploy)",
-            error: _
+            context: "with_payment_and_session",
+            error: _error_stirng
```
jacek-casper: same issue as above

lib/cli/simple_args.rs
```diff
@@ -344,7 +344,7 @@ fn split_arg(arg: &str) -> Result<(&str, &str, &str), CliError> {
 }
 
 /// Insert a value built from a single arg into `runtime_args`.
-pub(super) fn insert_arg(arg: &str, runtime_args: &mut RuntimeArgs) -> Result<(), CliError> {
+pub fn insert_arg(arg: &str, runtime_args: &mut RuntimeArgs) -> Result<(), CliError> {
```
jacek-casper, line 347: what's the reason for making this public

lib/cli/deploy.rs
```diff
@@ -230,7 +241,12 @@ pub fn make_transfer(
         payment_params,
         true,
     )?;
-    crate::output_deploy(output, &deploy).map_err(CliError::from)
+    #[cfg(feature = "std-fs-io")]
+    {
+        let output = parse::output_kind(maybe_output_path, force);
+        let _ = crate::output_deploy(output, &deploy).map_err(CliError::from);
```
jacek-casper:
```suggestion
        crate::output_deploy(output, &deploy).map_err(CliError::from)?;
```

lib/cli.rs
```diff
@@ -378,9 +381,18 @@ pub async fn get_balance(
...
+///
+/// # Parameters
+/// - `maybe_rpc_id`: The optional RPC ID as a string slice.
+/// - `node_address`: The address of the node as a string slice.
+/// - `verbosity_level`: The verbosity level as a 64-bit unsigned integer.
+/// - `maybe_block_id`: The optional block ID as a string slice.
+/// - `account_identifier`: The account identifier as a string slice.
+///
+/// # Returns
+/// - `Result<SuccessResponse<GetAccountResult>, CliError>`: The result containing either a successful response with the account details or a `CliError`.
```
jacek-casper: I don't think we need to name the actual type here
```suggestion
/// The result containing either a successful response with the account details or a `CliError`.
```

lib/cli/parse.rs
```diff
@@ -56,8 +65,12 @@ pub(super) fn secret_key_from_file<P: AsRef<Path>>(
 }
 
 pub(super) fn timestamp(value: &str) -> Result<Timestamp, CliError> {
+    #[cfg(any(feature = "std-fs-io", test))]
+    let timestamp = Timestamp::now();
+    #[cfg(not(any(feature = "std-fs-io", test)))]
+    let timestamp = Timestamp::zero();
```
jacek-casper, line 71: we should maybe not do any defaulting when the feature is off and just return an error? I think that might be less surprising

lib/cli/simple_args.rs
gRoussac: For the sdk using the lib to be able to add args in a a deploy and transaction

lib/cli/parse.rs
```diff
@@ -671,7 +734,7 @@ pub(super) fn global_state_identifier(
 }
 
 /// `purse_id` can be a formatted public key, account hash, or URef.  It may not be empty.
-pub(super) fn purse_identifier(purse_id: &str) -> Result<PurseIdentifier, CliError> {
+pub fn purse_identifier(purse_id: &str) -> Result<PurseIdentifier, CliError> {
```
jacek-casper, line 732: why are we making the identifier parsers public?

lib/cli/parse.rs
gRoussac: No that the purpose we want to avoid an error in that case when you are not in an OS but in wasm, being able to pass the step of construction with a zero timestamp because there is no clock in this context and the consumer can add a timestamp later with it's own clock and set_timestamp

lib/cli/simple_args.rs
jacek-casper: why not just use `insert_cl_value`? does it rely on the parsing code to convert an argument into cl type and value?

lib/cli/parse.rs
jacek-casper: we can just require them to pass the timestamp as an argument and only fail if they don't specify it

lib/cli/parse.rs
gRoussac: to be able to use it outside in query_balance and query_balance_details in the SDK with so that we avoid duplicating this parser helper like on dev account_identifier / transfer_target / args_from_simple_or_json / uref those helpers are very convenient to consume outside instead of recoding them again with same code 

```
use casper_client::cli::parse::purse_identifier as parse_purse_identifier;
```

jacek-casper: ok

lib/cli/simple_args.rs
gRoussac: I also use this 

```
/// Inserts an argument into a RuntimeArgs map.
///
/// # Arguments
///
/// * `args` - The RuntimeArgs map to insert the argument into.
/// * `new_arg` - The argument as a string.
///
/// # Returns
///
/// The modified `RuntimeArgs` map.
pub(crate) fn insert_arg(args: &mut RuntimeArgs, new_arg: String) -> &RuntimeArgs {
    match serde_json::from_str::<JsonArg>(&new_arg) {
        Ok(json_arg) => {
            if let Ok(named_arg) = NamedArg::try_from(json_arg.clone()) {
                // JSON args
                args.insert_cl_value(named_arg.name(), named_arg.cl_value().clone());
            }
        }
        Err(_) => {
            // Simple args
            let _ = casper_client::cli::insert_arg(&new_arg, args);
        }
    }
    args
}

File: lib/cli/parse.rs (hunk 12 above)
Comment (gRoussac) [reply lib/cli/parse.rs:71]: Sorry I made a confusion with this PR https://github.com/casper-network/casper-node/pull/4714/files#diff-28b84608019640a79626bab09ac38b648c6c87950af0d1f48c5bcd908533803bR99 in some constructors where you can not pass a value at that moment my bad.

So for this timestamp in create_transaction()

```
    let timestamp = parse::timestamp(transaction_params.timestamp)?;
```

I think we should not error when people are building a transaction from wasm because they miss a timestamp that can't be produced in wasm if they don't pass a value, I think the transaction should be with a zero timestamp and indeed they can add their time in transaction_params or they can call set_timestamp later but erroring here specifically for wasm for this timestamp is more confusing than letting it to zero without any indication in wasm.

Comment (jacek-casper) [reply lib/cli/parse.rs:71]: is there anything preventing the user from specifying the timestamp there? it seems odd to me to create it with a zero timestamp to just change it later

Comment (gRoussac) [reply lib/cli/parse.rs:71]: There is nothing for the consumer point of view to check that he passes a timestamp but it's even more surprising to fail at that moment if you don't pass it as not really compulsory in normal context, but you are in wasm we can't timestamp for you at that moment

in wasm 
```
const transaction_params = new TransactionStrParams(chain_name, initiator_addr, secret_key);
```
it would not be very  intuitive to fail here because timestamp is not given, it can be ensured by the SDK behind the hood you are right as a consumer in the SDK I ensure that the timestamp is there in params 

https://github.com/casper-ecosystem/rustSDK/blob/dev/src/types/transaction_params/transaction_str_params.rs#L276-L278

however if in the future somebody is using this builder from the client library like the SDK thus only in wasm transaction_params would not error, it would "just" have a zero timestamp and then if put_transaction this would be rejected so he would also have to add in his framework an 'auto' default timestamp from his wasm clock (chrono probably) if not there.

**Comment by gRoussac** [comment]
@jacek-casper I believe there is no CI/CD running on feat-track-2.0 branch and that we sould probably add it in  

https://github.com/casper-ecosystem/casper-client-rs/blob/rustSDK-feat-2.0/.github/workflows/ci-casper-client-rs.yml#L11

for this branch for PR because it seems to me the branch might not pass cargo audit and clippy

File: lib/rpcs/v2_0_0/get_entity.rs (hunk 4 above)
Comment (gRoussac) [reply lib/rpcs/v2_0_0/get_entity.rs]: thank you very much good point

File: lib/cli/transaction.rs (hunk 6 above)
Comment (gRoussac) [reply lib/cli/transaction.rs]: Thanks should be fixed in 

https://github.com/casper-ecosystem/casper-client-rs/pull/153/commits/b4a8f436f6d31c6cc38a1ad93843d9523279dd2c

File: lib/cli/tests.rs (hunk 7 above)
Comment (gRoussac) [reply lib/cli/tests.rs]: Thanks good catch I don't remember why I changed that probably linting. If this creates a binding instead of what I meant indeed it's always true. I think my eyes probably read assert_eq! which I use more often which would have kind of worked I guess with Partial Eq. I see I can't fool you with a badly used matches.

File: lib/cli/deploy.rs (hunk 10 above)
Comment (gRoussac) [reply lib/cli/deploy.rs]: Thanks should be fixed in

https://github.com/casper-ecosystem/casper-client-rs/commit/b4a8f436f6d31c6cc38a1ad93843d9523279dd2c

File: lib/cli.rs (hunk 11 above)
Comment (gRoussac) [reply lib/cli.rs]: Changed thanks

File: lib/cli/tests.rs (hunk 8 above)
Comment (gRoussac) [reply lib/cli/tests.rs]: plus a typo!

**Comment by jacek-casper** [comment]
You're right @gRoussac , the CI is not enabled here, it might be worth adding the branch to the workflow file:
https://github.com/casper-ecosystem/casper-client-rs/blob/feat-track-node-2.0/.github/workflows/ci-casper-client-rs.yml#L11

**Review by jacek-casper — APPROVED** [review]
lgtm to me, but I think it'd be good to get 1 more approval from someone else familiar with the client

**Comment by gRoussac** [comment]
@jacek-casper You review was great thanks again, I need the PR on node to be merged first and then I will imply Rafal for this one two before merging.

**Comment by gRoussac** [comment]
These PR needs to be backported on feat-track-2.0 and CI/CD implemented on branch

https://github.com/casper-ecosystem/casper-client-rs/pull/149
https://github.com/casper-ecosystem/casper-client-rs/pull/175

File: lib/cli/transaction.rs [hunk 14]
```diff
@@ -325,3 +332,49 @@ pub fn make_transaction_builder(
         }
     }
 }
+
+/// Retrieves a `SecretKey` based on the provided secret key string and configuration options.
+///
+/// # Arguments
+///
+/// * `secret_key` - A string representing the secret key. If empty, a `None` option is returned.
```
rafal-ch: It could also return an error in case `allow_unsigned_deploy` is `false`.

lib/cli/transaction.rs
```diff
@@ -325,3 +332,49 @@ pub fn make_transaction_builder(
...
+        Err(CliError::InvalidArgument {
+            context,
+            error: format!(
+                "allow_unsigned_deploy was {}, but no secret key was provided",
+                allow_unsigned_deploy
+            ),
+        })
+    } else {
+        Ok(None)
+    }
```
rafal-ch: A nit, but it might be more legible if we use `match (secret_key.is_empty(), allow_unsigned_deploy) {..}` statement here to avoid `if/else if/else` tree.

lib/cli/error.rs
```diff
@@ -8,7 +8,7 @@ use casper_types::{
     account::AccountHash, Key, NamedArg, PublicKey, RuntimeArgs, TimeDiff, Timestamp, URef,
 };
 use casper_types::{CLValueError, KeyFromStrError, UIntParseError, URefFromStrError};
-use uint::FromDecStrErr;
+pub use uint::FromDecStrErr;
```
rafal-ch, line 11: Can we leave this private and remove the reference to `FromDecStrErr` from `cli.rs`? We should be good, no?

rafal-ch (approved):
LGTM :+1: 
with just a few nits worth double checking

lib/cli/error.rs
gRoussac: I expose this error to the SDK like CliError, I would prefer to keep it please because right now I did not really clean on that side in the SDK, it has to import those errors right now.

```
use casper_client::{
    cli::JsonArgsError,
    cli::{CliError, FromDecStrErr},
    Error as CasperClientError,
};
```

lib/cli/transaction.rs
gRoussac: This commit should follow your reco 

https://github.com/casper-ecosystem/casper-client-rs/pull/153/commits/49cf58939cb148def454b9ebe5a67e1730e7e9b6

lib/cli/transaction.rs
gRoussac: I believe you meant to get clearer comment ? Please check this commit 

https://github.com/casper-ecosystem/casper-client-rs/pull/153/commits/58a301319d38a200832944d1b8be529f427fb08b

thanks

rafal-ch: yes, zactly

lib/cli/error.rs
rafal-ch: fair enough, thanks

PR #174, opened 2024-06-18
Add `CASPER_NODE_ADDRESS` env var.

mpapierski:
This makes it easier to use multiple commands without remembering to set `--node-address`. With this change, casper-client will try to load its value from env variable `CASPER_NODE_ADDRESS`.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -47,12 +47,13 @@ pub mod node_address {
     const ARG_VALUE_NAME: &str = "HOST:PORT";
     const ARG_DEFAULT: &str = "http://localhost:7777";
     const ARG_HELP: &str = "Hostname or IP and port of node on which HTTP service is running";
+    const ARG_ENV: &str = "CASPER_NODE_ADDRESS";
 
     pub fn arg(order: usize) -> Arg {
         Arg::new(ARG_NAME)
             .long(ARG_NAME)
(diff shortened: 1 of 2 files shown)


src/common.rs
```diff
@@ -47,11 +47,13 @@ pub mod node_address {
     const ARG_VALUE_NAME: &str = "HOST:PORT";
     const ARG_DEFAULT: &str = "http://localhost:7777";
     const ARG_HELP: &str = "Hostname or IP and port of node on which HTTP service is running";
+    const ARG_ENV: &str = "CASPER_NODE_ADDRESS";
```
gRoussac, line 50: Hi,

Can we try to avoid CASPER_ prefixes ? Is it still a node or an rpc RPC_ADDRESS ?

It would be great to unify that with CI/CD like

NODE_ADDRESS=http://localhost:7777 (or RPC_ADDRESS)
EVENT_ADDRESS=http://localhost:9999/events/main
SPECULATIVE_ADDRESS=http://localhost:7778

rafal-ch (approved):
LGTM :+1:

src/common.rs
gRoussac: to me the client should now use RPC_ADDRESS and the **binary client** NODE_ADDRESS

PR #176, opened 2024-06-18
Fix typo on pricing_mode and define pricing_mode ARG_VALUE_NAME options

gRoussac:
Fix typo on pricing_mode and define pricing_mode ARG_VALUE_NAME options

### DIFF
--- a/src/transaction/creation_common.rs
+++ b/src/transaction/creation_common.rs
@@ -433,18 +433,65 @@ pub(super) mod transfer_amount {
 
 pub(super) mod pricing_mode {
     use super::*;
+    use clap::{builder::PossibleValue, value_parser, ValueEnum};
+    use std::str::FromStr;
+
     pub(in crate::transaction) const ARG_NAME: &str = "pricing-mode";
 
-    const ARG_VALUE_NAME: &str = common::ARG_STRING;
+    const ARG_VALUE_NAME: &str = "classic|reserved|fixed";
     const ARG_HELP: &str = "Used to identify the payment mode chosen to execute the transaction";
+    const ARG_DEFAULT: &str = PricingMode::FIXED;
 
     pub(in crate::transaction) fn arg() -> Arg {
         Arg::new(ARG_NAME)
             .long(ARG_NAME)
             .required(false)
             .value_name(ARG_VALUE_NAME)
+            .default_value(ARG_DEFAULT)
             .help(ARG_HELP)
             .display_order(DisplayOrder::PricingMode as usize)
+            .value_parser(value_parser!(PricingMode))
+    }
+
+    #[derive(Debug, Clone, Copy)]
+    pub(super) enum PricingMode {
+        Classic,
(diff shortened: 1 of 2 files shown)


src/transaction/creation_common.rs
```diff
@@ -433,18 +433,65 @@ pub(super) mod transfer_amount {
...
+            match s.to_lowercase().as_str() {
+                PricingMode::CLASSIC => Ok(Self::Classic),
+                PricingMode::RESERVED => Ok(Self::Reserved),
+                PricingMode::FIXED => Ok(Self::Fixed),
+                _ => Err(format!("'{}' is not a valid pricing option", s)),
+            }
+        }
     }
 
     pub fn get(matches: &ArgMatches) -> &str {
```
jacek-casper: what's the purpose of defining the enum above - we do not seem to be using it to actually parse the argument?

gRoussac: Hi, to be able to use it from the library, it's exposing creation values, so that you can have the possible values, otherwise it's free text (and it's a bit hard to construct ``ARG_VALUE_NAME: &str = "classic|reserved|fixed";` from those enums) a bit like 
```
    pub(super) enum Category {
        InstallUpgrade,
        Large,
        Medium,
        Small,
    }
    
``` 
consumers can have defined enums for possible values when trying to building a transaction.

Desktop-screenshot - 2024-07-10T173132 217

jacek-casper: the enum is `pub(super)` though which means it won't be accessible outside of the crate

gRoussac: Correct sorry again I made a confusion with types. Indeed I would like those enums to be exported in fine, in a second PR maybe. I think it's not necessary for a consumer to have to redefine those like https://github.com/casper-ecosystem/rustSDK/blob/dev/src/types/pricing_mode.rs#L5-L18

Going back to your initial question I think you ask why is enum PricingMode defined ? then I think it's only for the line 

```.value_parser(value_parser!(PricingMode))```

I think I really duplicated the same code as TransactionCategory to have consistency between those two arguments building transactions. In a second time indeed I think they should ideally be moved to types and exported. I know the PricingMode in types and TransactionCategory are slightly different from those in the client but if they have to stay in the client, then I aim to export them from those crate mods to be able to re export them to consume outside so that also values are cascading, if they evolve in the client they have to evolve in consumer.

jacek-casper: I think this code might not work because the two types don't match, you create a value_parser for PricingMode but then retrieve it with `get_one::<String>`, might be worth checking manually if this works

gRoussac: good catch, bad copy paste from me 

Please check 

https://github.com/casper-ecosystem/casper-client-rs/pull/176/commits/08b90400f46e7000319412988353995088a2f9d6

jacek-casper (approved):

PR #178, opened 2024-06-24
Do not require payment amount for all transactions

jacek-casper:
We should only require it for classic pricing mode


alsrdn (approved):

rafal-ch (approved):
LGTM :+1:

PR #179, opened 2024-06-27
Add support for reward requests

### LINKED ISSUE #177: Add info_get_reward rpc
Add info_get_reward rpc

jacek-casper:
Confirmed to work via manual tests

### DIFF
--- a/src/get_reward.rs
+++ b/src/get_reward.rs
@@ -1,0 +1,144 @@
+use std::{fs, str};
+
+use async_trait::async_trait;
+use clap::{Arg, ArgMatches, Command};
+
+use casper_client::cli::CliError;
+use casper_types::PublicKey;
+
+use crate::{command::ClientCommand, common, Success};
+
+pub struct GetReward;
+
+/// This struct defines the order in which the args are shown for this subcommand's help message.
+enum DisplayOrder {
+    Verbose,
+    NodeAddress,
+    RpcId,
+    EraIdentifier,
+    Validator,
+    Delegator,
+}
+
+/// Handles providing the arg for and retrieval of public keys.
+mod key {
+    use casper_types::AsymmetricType;
+
+    use super::*;
+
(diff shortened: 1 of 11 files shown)


src/get_reward.rs
```diff
@@ -0,0 +1,144 @@
...
+
+mod validator {
+    use super::*;
+
+    pub(crate) const ARG_NAME: &str = "validator";
+    const ARG_HELP: &str =
+        "A public key of the validator, formatted as a hex-encoded string or a path to a file";
+
+    pub(super) fn arg() -> Arg {
+        key::arg(ARG_NAME, ARG_HELP, DisplayOrder::Validator as usize)
```
rafal-ch: Should validator arg be set to `required(true)`?

src/get_reward.rs
```diff
@@ -0,0 +1,144 @@
...
+        Ok(value.to_string())
+    }
+}
+
+mod validator {
+    use super::*;
+
+    pub(crate) const ARG_NAME: &str = "validator";
+    const ARG_HELP: &str =
+        "A public key of the validator, formatted as a hex-encoded string or a path to a file";
```
rafal-ch, line 75: While the comment says that it could be a path to a file, in the `get_reward()` it is immediately parsed as `PublicKey`.
Please make sure that we can indeed read the key from a provided file to stay consistent with other commands.

jacek-casper: this does support reading from files, we call `key::get(ARG_NAME, matches)` on line 83, which attempts to read it from a file (I've used it and it does work with a path to a file)

src/get_reward.rs
jacek-casper: that's true, we can set it to required 👍

src/get_reward.rs
rafal-ch: ok, thanks for double checking that

rafal-ch (approved):
LGTM :+1:

PR #175, opened 2024-06-18
#163 Handle git dependency in build.rs

gRoussac:
If the git commit hash is retrieved successfully, set the environment variable

PR #185, opened 2024-07-17
Remove vergen dep on feat-track-2.0 for cargo audit 

### LINKED ISSUE #186: CI/CD on feat-track-2.0
CI/CD on feat-track-2.0 should pass on this branch

gRoussac:
We need to have Ci/CD and cargo audit passing on feat-track-2.0. I suggest we cherry-pick some changes from dev to remove vergen, prepare code verification changes and for now still bypass "std-fs-io" feature

- Port from dev to feat-track-2.0 of #149 #175 and partially #144 to remove vergen and set tokio as dep
- clippy/test/build with no default features not yet working until #153 and https://github.com/casper-network/casper-node/pull/4714 are merged

### DIFF
--- a/.github/workflows/ci-casper-client-rs.yml
+++ b/.github/workflows/ci-casper-client-rs.yml
@@ -53,12 +53,11 @@ jobs:
           command: clippy
           args: --all-targets
 
-      # // Awaits https://github.com/casper-ecosystem/casper-client-rs/pull/153
-      # - name: Clippy with no features
-      #   uses: actions-rs/cargo@v1
-      #   with:
-      #     command: clippy
-      #     args: --all-targets --no-default-features
+      - name: Clippy with no features
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --all-targets
 
       - name: Doc
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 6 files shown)


rafal-ch (approved):
LGTM :+1:

.github/workflows/ci-casper-client-rs.yml
```diff
@@ -45,6 +53,13 @@ jobs:
           command: clippy
           args: --all-targets
 
+      # // Awaits https://github.com/casper-ecosystem/casper-client-rs/pull/153
+      # - name: Clippy with no features
+      #   uses: actions-rs/cargo@v1
+      #   with:
+      #     command: clippy
+      #     args: --all-targets --no-default-features
```
jacek-casper: I think we should not comment out these commands, just include them without the args for now, and add `--all-targets --no-default-features` in the wasm PR

.github/workflows/ci-casper-client-rs.yml
```diff
@@ -55,3 +70,17 @@ jobs:
...
+      #   with:
+      #     command: test
+      #     args: --no-default-features
+
+      # // Awaits https://github.com/casper-network/casper-node/pull/4714
+      # - name: Build lib for Wasm with no features
+      #   uses: actions-rs/cargo@v1
+      #   with:
+      #     command: build
+      #     args: --lib --target wasm32-unknown-unknown --no-default-features
```
jacek-casper: same as above

.github/workflows/ci-casper-client-rs.yml
gRoussac: Ok but those are already on dev branch https://github.com/casper-ecosystem/casper-client-rs/blob/dev/.github/workflows/ci-casper-client-rs.yml

and the feature it self is already on feat-track-2.0, it will still fail until the PR is merged so I commented out.

it is more confusing to repeat the every command twice and change dev version which is correct.

the version uncommitted is still not correct imo. https://github.com/casper-ecosystem/casper-client-rs/pull/185/commits/bd9d4f591a0cdafdc538f5e3b700e72296ff04b6

jacek-casper: the commands pass on my machine

jacek-casper (approved):

PR #189, opened 2024-07-22
Update for changed interface of 'pricing_mode' enum

rafal-ch:
Recently in [this PR](https://github.com/casper-ecosystem/casper-client-rs/pull/176), we changed the signature of `PricingMode::get()` from

`pub fn get(matches: &ArgMatches) -> &str`

to

`pub fn get(matches: &ArgMatches) -> Option<&PricingMode>`

This broke the compatibility with `TransactionStrParams` which expects `&str`.

The fix in this PR patches this, which should fix the node builds on CI. However, we might reconsider if the aforementioned change in the interface should be reverted.

### DIFF
--- a/src/transaction/creation_common.rs
+++ b/src/transaction/creation_common.rs
@@ -454,7 +454,7 @@ pub(super) mod pricing_mode {
     }
 
     #[derive(Debug, Clone, Copy)]
-    pub(super) enum PricingMode {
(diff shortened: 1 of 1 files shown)


jacek-casper (approved):

gRoussac:
My bad, sorry for that

Issue #198, opened 2024-11-13
Add `put-transaction` subcommands for managing delegator reservations


wojcik91:
Add `add-reservations` and `cancel-reservations` subcommands for interactions with new Auction entrypoints.

Note: Add dedicated CLI options for passing in reservations and pass them along to the node as CLType::Any.

devendran-m:
Assign it to Maciej

wojcik91:
Relevant updates to casper-types were implemented in https://github.com/casper-network/casper-node/pull/4971

Issue #203, opened 2024-11-25
Implement `reserved_slots` in casper-client-rs


devendran-m:
Ticket to implement `reserved_slots` in `casper-client-rs`

PR #207, opened 2024-11-27
Add support for new auction entrypoints

wojcik91:
- fix `reserved_slots` handling in `add_bid`
- add new subcommands for `put-transaction` and `make-transaction`:
  - `change_bid_public_key`
  - 'add_reservations`
  - `cancel_reservations` 

Closes https://github.com/casper-ecosystem/casper-client-rs/issues/198


devendran-m:
Also closes #203

public_key.pem
zajko, line 1: Were these committed intentionally?

Cargo.toml
```diff
@@ -57,6 +57,7 @@ uint = "0.9.5"
 
 [dev-dependencies]
 tempfile = "3.8.1"
+casper-types = { version = "5.0.0", features = ["std", "json-schema", "testing"] }
```
zajko, line 60: Shouldn't we be pointing to `feat-2.0` ?

public_key.pem
wojcik91: Nope. Sorry about that

Cargo.toml
wojcik91: Yes, it's configured below.

zajko (approved):

PR #208, opened 2024-11-27
Making transferred-value optional in the CLI. Also, if VM1 is the sel…

zajko:
…ected (or defaulted) runtime, `--transferred-value` argument is not allowed. This doesn't 100% reflect the API, since the API actually allows to provide a value for that field in case of VM1, but to limit confusion we reject it in the CLI since it has no meaning for VM1

### DIFF
--- a/src/transaction/creation_common.rs
+++ b/src/transaction/creation_common.rs
@@ -871,8 +871,22 @@ pub(super) mod transferred_value {
             .display_order(DisplayOrder::TransferredValue as usize)
     }
 
-    pub(super) fn get(matches: &ArgMatches) -> Option<&String> {
+    pub(super) fn get_raw(matches: &ArgMatches) -> Option<&String> {
         matches.get_one(ARG_NAME)
+    }
+
+    pub(super) fn get(matches: &ArgMatches) -> Result<Option<u64>, CliError> {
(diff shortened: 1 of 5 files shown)


src/transaction/creation_common.rs
```diff
@@ -1611,14 +1612,15 @@ pub(super) mod invocable_entity {
         let runtime = transaction_runtime::get(matches);
 
         let entry_point = session_entry_point::get(matches).unwrap_or_default();
-        let transferred_value = transferred_value::get(matches)
-            .map(|value| value.parse::<u64>().unwrap())
-            .unwrap_or_default();
+        let maybe_transferred_value =
+            transferred_value::get(matches).map(|value| value.parse::<u64>().unwrap());
```
wojcik91: We should remove this unwrap and parse it similarly to other args in case someone provides an invalid string.

Maybe add a `parse` method that would also handle the validation below.

mpapierski: Agreed. How about this magic one-liner:
```suggestion
            transferred_value::get(matches).and_then(|value| value.parse::<u64>().ok());
```

mpapierski: But probably better to add an error variant to report invalid integer rather than panic (or ignore it as in the one-liner above)

mpapierski (approved):

PR #220, opened 2024-12-11
Pub DeployBuilder

gRoussac:
This helper was public in casper node and consumed by crates that can build deploy easily. Would you please consider making it public ? Related to https://github.com/casper-ecosystem/casper-client-rs/pull/212


zajko:
hi @gRoussac 
This is a valid request - just one thing, what's your use case? Do you require only DeployBuilder or do you want also the TransactionV1Builder available from casper-client-rs.

zajko (approved):

PR #226, opened 2025-01-22
Add contract package arg

darthsiroftardis:
CHANGELOG:

- Added `contract-hash` as an optional arg for `invocable-entity` which can be passed instead of an entity addr as any `hash-` prefixes hex encoded string
- Added `contract-package-hash` as an optional arg for `package` which can be passed instead of an entity addr as any `hash-` prefixes hex encoded string

### DIFF
--- a/src/transaction/creation_common.rs
+++ b/src/transaction/creation_common.rs
@@ -1021,10 +1021,10 @@ pub(super) mod entity_addr {
 }
 
 pub(super) mod contract_hash {
-    use casper_types::{EntityAddr, Key};
     use super::*;
     use casper_client::cli::CliError;
     use casper_client::Error;
+    use casper_types::{EntityAddr, Key};
 
     pub const ARG_NAME: &str = "contract-hash";
     const ARG_VALUE_NAME: &str = "FORMATTED STRING";
@@ -1039,9 +1039,7 @@ pub(super) mod contract_hash {
     }
 
     pub fn get(matches: &ArgMatches) -> Option<&str> {
-        matches
-            .get_one::<String>(ARG_NAME)
-            .map(String::as_str)
+        matches.get_one::<String>(ARG_NAME).map(String::as_str)
     }
 
     pub(super) fn parse_contract_hash(value: &str) -> Result<EntityAddr, CliError> {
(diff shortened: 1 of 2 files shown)


src/transaction/creation_common.rs
```diff
@@ -1012,7 +1013,45 @@ pub(super) mod entity_addr {
...
+
+pub(super) mod contract_hash {
+    use super::*;
+    use casper_client::cli::CliError;
+    use casper_client::Error;
+    use casper_types::{EntityAddr, Key};
+
+    pub const ARG_NAME: &str = "contract-hash";
+    const ARG_VALUE_NAME: &str = "FORMATTED STRING";
+    const ARG_HELP: &str = "The formatted string prefixed by hash- representing a 1.x contract.";
```
davidatwhiletrue, line 1031: only 1.x contracts? or any contract?

src/transaction/creation_common.rs
```diff
@@ -1062,6 +1100,44 @@ pub(super) mod package_addr {
...
 
+pub(super) mod contract_package_hash {
+    use super::*;
+    use casper_client::{cli::CliError, Error};
+    use casper_types::{Key, PackageAddr};
+
+    pub const ARG_NAME: &str = "contract-package-hash";
+    const ARG_VALUE_NAME: &str = "FORMATTED STRING";
+    const ARG_HELP: &str =
+        "The formatted string prefixed by hash- representing a 1.x contract package";
```
davidatwhiletrue, line 1111: Only 1.x package? or any package?

davidatwhiletrue (commented):
@darthsiroftardis see my comment

src/transaction/creation_common.rs
darthsiroftardis: Yeah, its any package

darthsiroftardis: I can amend the help statement

src/transaction/creation_common.rs
darthsiroftardis: Any contract, addressed in 1.x scheme

zajko (approved):

Issue #225, opened 2025-01-22
Update `list-deploys` subcommand for Condor if necessary


wojcik91:
Verify if `list-deploys` subcommand is compatible with new blocks and update it if necessary.

Also add an equivalent `list-transactions` subcommand.

PR #228, opened 2025-01-23
Add `list-transactions` subcommand

wojcik91:
Also add `node_address` parsing to avoid opaque errors and add a request timeout.

Closes https://github.com/casper-ecosystem/casper-client-rs/issues/225

### DIFF
--- a/lib/json_rpc/call.rs
+++ b/lib/json_rpc/call.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use crate::{Error, JsonRpcId, SuccessResponse, Verbosity};
 use jsonrpc_lite::{JsonRpc, Params};
 use once_cell::sync::OnceCell;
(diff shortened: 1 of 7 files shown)


lib/json_rpc/call.rs
```diff
@@ -56,9 +66,14 @@ impl Call {
         #[cfg(feature = "std-fs-io")]
         crate::json_pretty_print(&rpc_request, self.verbosity)?;
 
-        let client = CLIENT.get_or_init(Client::new);
+        let client = CLIENT.get_or_init(|| {
+            Client::builder()
+                .timeout(REQUEST_TIMEOUT)
```
zajko: timeout should probably be an optional argument for any of the commands which defaults to REQUEST_TIMEOUT.
We don't have to do that in this PR IMHO

zajko (approved):

PR #250, opened 2025-05-20
Add new subcommand `withdraw-bid-all`

darthsiroftardis:
CHANGLEOG

- Adds a new subcommand `withdraw-bid-all` which takes in a public-key for a validator and produces a withdraw bid transaction which completely unstakes the validator
- Modified `withdraw-bid` to check that withdraw bid does not result in a validator's stake dropping below a validator minimum bid threshold
- Added an override flag to `withdraw-bid` to simply log the warning to stdout and produce/send the withdraw bid transaction

### DIFF
--- a/lib/cli/transaction_builder_params.rs
+++ b/lib/cli/transaction_builder_params.rs
@@ -1,4 +1,9 @@
-use casper_types::{bytesrepr::Bytes, system::auction::{DelegatorKind, Reservation}, AddressableEntityHash, PackageHash, PublicKey, TransactionRuntimeParams, TransferTarget, URef, U512, EntityVersionKey};
+use casper_types::{
+    bytesrepr::Bytes,
+    system::auction::{DelegatorKind, Reservation},
+    AddressableEntityHash, EntityVersionKey, PackageHash, PublicKey, TransactionRuntimeParams,
+    TransferTarget, URef, U512,
+};
 
 /// An enum representing the parameters needed to construct a transaction builder
 /// for the commands concerning the creation of a transaction
--- a/lib/cli/transaction_builder_params.rs
+++ b/lib/cli/transaction_builder_params.rs
@@ -139,6 +139,8 @@ pub enum TransactionBuilderParams<'a> {
         public_key: PublicKey,
         /// The amount to be withdrawn in the withdraw bid transaction
         amount: U512,
+        /// Override the min bid amount check
+        min_bid_override: bool,
     },
     /// Parameters for the activate bid variant of the transaction builder
     ActivateBid {
--- a/lib/cli/transaction_builder_params.rs
+++ b/lib/cli/transaction_builder_params.rs
@@ -1,9 +1,4 @@
-use casper_types::{
-    bytesrepr::Bytes,
-    system::auction::{DelegatorKind, Reservation},
-    AddressableEntityHash, PackageHash, PublicKey, TransactionRuntimeParams, TransferTarget, URef,
-    U512,
-};
+use casper_types::{bytesrepr::Bytes, system::auction::{DelegatorKind, Reservation}, AddressableEntityHash, PackageHash, PublicKey, TransactionRuntimeParams, TransferTarget, URef, U512, EntityVersionKey};
 
 /// An enum representing the parameters needed to construct a transaction builder
 /// for the commands concerning the creation of a transaction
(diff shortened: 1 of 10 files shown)


lib/cli/transaction_builder_params.rs
```diff
@@ -139,6 +161,8 @@ pub enum TransactionBuilderParams<'a> {
         public_key: PublicKey,
         /// The amount to be withdrawn in the withdraw bid transaction
         amount: U512,
+        /// Override the min bid amount check
+        min_bid_override: bool,
```
zajko, line 165: Maybe it's the language barrier, but I'm not sure how to interpret this flag - will the client enforce the check if this is false or will the client enforce the check if this is true?

src/transaction/creation_common.rs
```diff
@@ -2231,12 +2345,21 @@ pub(super) mod package_alias {
 
         let maybe_entity_version = session_version::get(matches);
 
+        let maybe_entity_version_key = if let Some(entity_version) = maybe_entity_version {
+            match major_version::get(matches) {
+                Some(major) => Some(EntityVersionKey::new(major, entity_version)),
```
zajko, line 2373: Not sure if it's an issue but do we care the other way around? If someone specifies major but not entity_version? major would be silently ignored from what I can see. We can also handle this in another PR

src/transaction/put.rs
```diff
@@ -45,56 +82,367 @@ impl ClientCommand for PutTransaction {
...
-                invocable_entity_alias::NAME => {
-                    parse_rpc_args_and_run(matches, invocable_entity_alias::run)?
-                }
-                package::NAME => parse_rpc_args_and_run(matches, package::run)?,
-                package_alias::NAME => parse_rpc_args_and_run(matches, package_alias::run)?,
-                session::NAME => parse_rpc_args_and_run(matches, session::run)?,
-                transfer::NAME => parse_rpc_args_and_run(matches, transfer::run)?,
+        match matches.subcommand() {
+            None => Err(CliError::InvalidArgument {
+                context: "Make Transaction",
```
zajko, line 87: `  context: "Put Transaction",`
?

src/transaction/put.rs
```diff
@@ -1,17 +1,53 @@
...
+    version: ProtocolVersion,
+    hard_reset: bool,
+    activation_point: ActivationPoint,
+}
+
+/// A chainspec configuration as laid out in the TOML-encoded configuration file.
+#[derive(PartialEq, Eq, Serialize, Deserialize, Debug)]
+// Disallow unknown fields to ensure config files and command-line overrides contain valid keys.
+#[serde(deny_unknown_fields)]
+pub(super) struct TomlChainspec {
```
zajko, line 40: Do we really need all these fields and the `deny_unknown_fields` property? If there is a new property in the chainspec that the client doesn't care about we should automatically crash?

src/transaction/put.rs
```diff
@@ -45,56 +82,367 @@ impl ClientCommand for PutTransaction {
...
+
+    let (transaction_builder_params, transaction_str_params) = withdraw_bid::run(matches)?;
+
+    if let TransactionBuilderParams::WithdrawBid {
+        public_key,
+        amount,
+        min_bid_override,
+    } = &transaction_builder_params
+    {
+        let chainspec_bytes = casper_client::cli::get_chainspec("", node_address, verbosity_level)
```
zajko, line 199: We don't cache for now?

src/transaction/put.rs
```diff
@@ -45,56 +82,367 @@ impl ClientCommand for PutTransaction {
...
+            .await?
+            .result
+            .auction_state
+            .bids()
+            .find(|(bid_key, _bid)| **bid_key == *public_key)
+        {
+            Some((_, bid)) => {
+                let staked_amount = *bid.staked_amount();
+                let remainder = staked_amount.saturating_sub(*amount);
+                if remainder < U512::from(minimum_validator_bid) {
```
zajko, line 219: We should amend the description of the withdraw bid command of the client to make the users aware that this is a "best effort" guardrail. It will deflect 95% of erroneous behavior, but ultimately it is not enforced on the node so due to race conditions it can still allow a unbid below minimum bid even if the flag is `false`

zajko (approved):

src/transaction/creation_common.rs
darthsiroftardis: The issue, is that when specifying a version, we need both parts of the info. So yes, I do agree with the approach!

src/transaction/put.rs
darthsiroftardis: Nope, right now its completely stateless

src/transaction/put.rs
darthsiroftardis: I can add more detail

lib/cli/transaction_builder_params.rs
darthsiroftardis: The client will enforce the check if the flag isn't provided

src/transaction/put.rs
darthsiroftardis: We probably don't

PR #258, opened 2025-07-21
Additional guardrails for withdraw bid

darthsiroftardis:
CHANGELOG:

- Extended withdraw bid guardrails to `put-transaction` and `put-deploy` subcommands that invoke the entry point via stored contract by hash/name and package by hash/name


lib/cli/deploy.rs
```diff
@@ -17,6 +20,289 @@ use crate::{
...
+struct TomlProtocol {
+    version: ProtocolVersion,
+    hard_reset: bool,
+    activation_point: ActivationPoint,
+}
+
+/// A chainspec configuration as laid out in the TOML-encoded configuration file.
+#[derive(PartialEq, Eq, Serialize, Deserialize, Debug)]
+// Disallow unknown fields to ensure config files and command-line overrides contain valid keys.
+#[serde(deny_unknown_fields)]
```
zajko, line 43: Why do we deny unknown fields? wouldn't it be easier to only define the fields we want and let it be a forward-compatible permissive structure?

lib/cli/deploy.rs
```diff
@@ -17,6 +20,289 @@ use crate::{
...
+async fn check_auction_state_for_withdraw(
+    node_address: &str,
+    verbosity_level: u64,
+    hash_addr: HashAddr,
+    entry_point_name: String,
+    runtime_args: &RuntimeArgs,
+    min_bid_override: bool,
+) -> Result<(), CliError> {
+    // Best guess on the entry point name
+    if entry_point_name == *"withdraw_bid" {
```
zajko, line 108: would it make sense to have this as the default but allow the user to give an override?

lib/cli/transaction.rs
```diff
@@ -455,3 +460,211 @@ pub fn get_maybe_secret_key(
...
+                            CliError::InvalidCLValue("failed to get public key arg".to_string())
+                        })?
+                        .to_t::<PublicKey>()
+                        .map_err(|err| CliError::InvalidCLValue(err.to_string()))?;
+
+                    do_withdraw_amount_checks(node_address, 0, public_key, amount, min_bid_override)
+                        .await?
+                }
+            } else {
+                println!("Skipping amount checks for withdraw bid")
```
zajko, line 665: Maybe this should be an info log?

lib/cli/deploy.rs
```diff
@@ -17,6 +20,289 @@ use crate::{
...
+                        0,
+                        hash_addr,
+                        entry_point.clone(),
+                        args,
+                        min_bid_override,
+                    )
+                    .await
+                }
+                None => {
+                    println!("unable to get named key skipping withdrawal checks");
```
zajko, line 236: Maybe this should be an info log?

lib/cli/deploy.rs
```diff
@@ -17,6 +20,289 @@ use crate::{
...
```
zajko, line 298: Maybe this should be an info log?

lib/cli/deploy.rs
```diff
@@ -17,6 +20,289 @@ use crate::{
...
+        .find(|(bid_key, _bid)| **bid_key == public_key)
+    {
+        Some((_, bid)) => {
+            let staked_amount = *bid.staked_amount();
+            let remainder = staked_amount.saturating_sub(amount);
+            if remainder < U512::from(minimum_validator_bid) {
+                if !min_bid_override {
+                    return Err(CliError::ReducedStakeBelowMinAmount);
+                } else {
+                    println!("[WARN] Execution of this withdraw bid will result in unbonding of all stake")
```
zajko, line 89: Maybe this should be an info log?

lib/cli/transaction.rs
```diff
@@ -455,3 +460,211 @@ pub fn get_maybe_secret_key(
...
+    let state_root_hash = *get_block("", node_address, 0, "")
+        .await?
+        .result
+        .block_with_signatures
+        .ok_or_else(|| CliError::FailedToGetStateRootHash)?
+        .block
+        .state_root_hash();
+    let encoded_hash = base16::encode_lower(&state_root_hash);
+    match transaction {
+        Transaction::Deploy(_) => return Ok(()),
```
zajko, line 479: Why we don't do any checks in this variant?
Also, do we do these checks if we call `send-deploy` or `send-transaction`?

lib/cli/deploy.rs
darthsiroftardis: No, that wouldn't make sense, because we specifically follow up and check the invocation target is the auction hash. So it wouldnt make sense because there is only the one `withdraw_bid` entrypoint

lib/cli/transaction.rs
darthsiroftardis: We don't have logs in the client so this print is probably better to have to just give the user some awarness of whether the check is happening

lib/cli/transaction.rs
darthsiroftardis: Right, so we do the deploy checks just before we do `put_deploy` in a separate code path in the deploy mod

We do checks for both just not in this one place

lib/cli/deploy.rs
darthsiroftardis: We should def highlight this error more here and clearly call out the affect

lib/cli/transaction.rs
zajko: We don't have logs, but we have verbosity levels.
If someone uses the CLI in their automation they will have a hard time parsing these informational outputs. I see that the client often does sth like:
```
if verbosity == Verbosity::Medium || verbosity == Verbosity::High {
        println!("Hash: {hash_str}");
        println!("Verification service base path: {verification_url_base_path}",);
    }
```
I'm OK with having a spin-off ticket to address that.

lib/cli/transaction.rs
zajko: You will go through here if you use a transaction stored in json and do send-transaction AFAIK.
Still - if you don't expect to have a Transaction::Deploy in this flow I would error-out here, not return Ok(())

lib/cli/transaction.rs
darthsiroftardis: Sounds good!

zajko (approved):
There are a few things we need to follow up on, but we can do that in another ticket as we discussed.

PR #259, opened 2025-07-22
Optimistically deserialize toml chainspec

darthsiroftardis:
CHANGELOG:

- When attempting to deserialize the chainspec skip the minimum amount checks on failure instead of panicking

### DIFF
--- a/src/transaction/put.rs
+++ b/src/transaction/put.rs
@@ -196,49 +196,27 @@ async fn put_withdraw_bid_transaction(matches: &ArgMatches) -> Result<Success, C
         min_bid_override,
     } = &transaction_builder_params
     {
+        if *min_bid_override {
+            return casper_client::cli::put_transaction(
+                rpc_id,
+                node_address,
+                verbosity_level,
+                transaction_builder_params,
+                transaction_str_params,
+            )
+            .await
+            .map(Success::from);
+        }
+
         let chainspec_bytes = casper_client::cli::get_chainspec("", node_address, verbosity_level)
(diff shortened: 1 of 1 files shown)


igor-casper (approved):

src/transaction/put.rs
```diff
@@ -201,8 +201,36 @@ async fn put_withdraw_bid_transaction(matches: &ArgMatches) -> Result<Success, C
             .result
             .chainspec_bytes;
 
-        let chainspec_as_str = std::str::from_utf8(chainspec_bytes.chainspec_bytes()).unwrap();
-        let toml_chainspec: TomlChainspec = toml::from_str(chainspec_as_str).unwrap();
+        let chainspec_as_str = match std::str::from_utf8(chainspec_bytes.chainspec_bytes()) {
+            Ok(chainspec_as_str) => chainspec_as_str,
+            Err(_) => {
+                println!("Unable to decode chainspec bytes, skipping withdraw bid checks");
+                return casper_client::cli::put_transaction(
```
zajko: I'm not sure understand this assumption - The user asks to run withdraw bid with additional guardrails and we drop those guardrails if we can't deserialize chainspec? In my mind it would be better to error out with a message. If the user can't run withdraw bid with additional guardrails they should still be able to run without the checks, right?

darthsiroftardis: Yeah, I can sharpen this a bit to include the override and then error out accordingly, good catch

zajko (approved):

PR #263, opened 2025-08-25
Fix uneccessary print message for put-transaction

darthsiroftardis:
Returns early instead of printing a message in non-withdraw cases

### DIFF
--- a/.github/workflows/ci-casper-client-rs.yml
+++ b/.github/workflows/ci-casper-client-rs.yml
@@ -46,7 +46,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: audit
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-casper-client-rs.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: audit
-          args: --ignore RUSTSEC-2025-0009 --ignore RUSTSEC-2025-0022 --ignore RUSTSEC-2025-0023
+          args: --ignore RUSTSEC-2025-0009 --ignore RUSTSEC-2025-0022 --ignore RUSTSEC-2025-0023 --ignore RUSTSEC-2025-0047
```
zajko: if you do `cargo update` you won't need this ignore

zajko (approved):

